PR #3, opened 2017-01-20
Update Travis configuration + release documentation

sparuvu:

* Updated the release documentation
* Updated the travis configuration & scripts to build and deploy the tagged artifacts to maven central.
* Updated the site publish script to build and update the gh-pages branches automatically everytime the project is released (tagged).
 
How was it tested?
----
* Configured the travis temporarily build my branch.

How to test
----

Update the travis configuration to build a specific branch instead of master and push a commit. Travis should automatically start a build.

Reviewers
----


RELEASING.md
```diff
@@ -2,30 +2,50 @@
 
 ## Preparing for the Release
 
-* A valid Sonatype user account is required to release this project. To create an account sign up with [Sonatype](https://issues.sonatype.org/secure/Signup!default.jspa).
-* Install `gpg2` - we will be using this tool to automatically sign off our artifacts
-	* install it via brew - `brew install gpg2`. There are other ways to install this tool but doing it via brew can help us all be in sync easily with the version of the tool we are using.
-* Follow this [guide](http://central.sonatype.org/pages/working-with-pgp-signatures.html#generating-a-key-pair) to generate your own gpg key and secret.
-* Execute the following commands in your terminal to prepare your travis account to deploy to maven central
+* A valid Sonatype user account is required to release this project. To create an account sign up with [Sonatype](https://issues.sonatype.org/secure/Signup!default.jspa). Sonatype requires a valid email Id with the domain `cerner.com` to confirm that we are deploying the artifacts to the right namespace.
```
nimeshsubramanian: nit: ID

RELEASING.md
```diff
@@ -2,30 +2,50 @@
...
+	* **Note**: Make sure to use the cerner email to signup.
```
nimeshsubramanian: Nit: `Cerner`

RELEASING.md
```diff
@@ -47,12 +67,19 @@ After preparing the machine follow the below steps
...
+```
+mvn release:rollback
+```
 
-    ```
-    mvn release:rollback
-    ```
+Once the master branch and tags are pushed the following happens.
+
+* Travis kicks off a new build for released tag pushes the artifact to Maven central. It roughly takes about 2 hours for the artifacts to sync with the maven central.
```
nimeshsubramanian: NIT: can we rephrase the first sentence?
maybe: A new build is run by Travis to push the released artifact to Maven central.

johnlcox:
Are the `*.enc` files needed?

johnlcox:
Looks like the `.enc` files are used, but what is the `deploy_site.enc` file? I think I understand the GPG files, but I'm not sure what this deploy on is.

.travis.yml
```diff
@@ -1,15 +1,35 @@
...
+  - ".travis/publish.sh"
+  skip_cleanup: true
+  on:
+    repo: cerner/beadledom
+    tags: true
 env:
   global:
-  - GPG_DIR="`pwd`/deploy"
+  - GPG_DIR=${TRAVIS_BUILD_DIR}/.travis/
+  - ENCRYPTION_LABEL: 4d3aca009c62
```
sparuvu: should delete this

.travis.yml
```diff
@@ -1,15 +1,35 @@
...
+- oraclejdk8
 cache:
   directories:
   - "$HOME/.m2"
-after_success:
-- openssl aes-256-cbc -pass pass:$ENCRYPTION_PASSWORD -in $GPG_DIR/pubring.gpg.enc -out $GPG_DIR/pubring.gpg -d
-- openssl aes-256-cbc -pass pass:$ENCRYPTION_PASSWORD -in $GPG_DIR/secring.gpg.enc -out $GPG_DIR/secring.gpg -d
-- "$GPG_DIR/publish.sh"
+install:
+#- sudo apt-get install -y gnupg2
```
nimeshsubramanian: NIT: remove commented out line

sparuvu: I think we should add this back. I have seen others installing gpg2.

sparuvu:
> Looks like the .enc files are used, but what is the deploy_site.enc file? I think I understand the GPG files, but I'm not sure what this deploy on is.

deploy_site.enc is the encrypted private ssh key for travis to push the gh-pages branch. I have added the corresponding public key to beadledom's deploy keys section

.travis/publish.sh
```diff
@@ -0,0 +1,17 @@
...
+openssl aes-256-cbc -pass pass:$ENCRYPTION_PASSWORD -in $GPG_DIR/secring.gpg.enc -out $GPG_DIR/secring.gpg -d
+
+if [ ! -z "$TRAVIS_TAG" ]
+then
+    echo "Checking out a tag $TRAVIS_TAG"
+    git checkout tags/$TRAVIS_TAG -b tag_$TRAVIS_TAG
+    mvn deploy --settings $GPG_DIR/settings.xml -DskipTests=true -B -U
+    ${GPG_DIR}/publish_docs.sh $TRAVIS_TAG
+else
+    echo "not on a tag -> keep snapshot version in pom.xml"
```
johnlcox: Is this comment a legacy of some of those other examples that swapped the versions? Maybe this should just say something like `deploying SNAPSHOT from master`.

.travis/publish.sh
```diff
@@ -0,0 +1,17 @@
...
+echo "executing deploy.sh"
+echo "decrypting keyrings"
+openssl aes-256-cbc -pass pass:$ENCRYPTION_PASSWORD -in $GPG_DIR/pubring.gpg.enc -out $GPG_DIR/pubring.gpg -d
+openssl aes-256-cbc -pass pass:$ENCRYPTION_PASSWORD -in $GPG_DIR/secring.gpg.enc -out $GPG_DIR/secring.gpg -d
+
+if [ ! -z "$TRAVIS_TAG" ]
+then
+    echo "Checking out a tag $TRAVIS_TAG"
+    git checkout tags/$TRAVIS_TAG -b tag_$TRAVIS_TAG
+    mvn deploy --settings $GPG_DIR/settings.xml -DskipTests=true -B -U
```
johnlcox: If this build fails for an external reason, like a network issue, can we re-run the build without pushing another tag?

.travis/publish_docs.sh
```diff
@@ -19,9 +21,20 @@ BWHITE='\e[1;37m'       # Bold White
 
 function display_usage() {
   printf "${GREEN}Usage:\n"
-  printf "${YELLOW}      ./prepare_site.sh [version|dev]\n\n"
-  printf "${GREEN}   Example: ${YELLOW}./prepare_site.sh 2.2 ${GREEN}- Prepare docs for 2.2 release.\n"
-  printf "${GREEN}   Example: ${YELLOW}./prepare_site.sh dev ${GREEN}- Prepare docs for current SNAPSHOT.\n$RESET"
+  printf "${YELLOW}      ./publish_docs.sh [version|dev]\n\n"
```
johnlcox: Why `docs` instead of `site`?

.travis/publish_docs.sh
```diff
@@ -19,9 +21,20 @@ BWHITE='\e[1;37m'       # Bold White
...
+  printf "${GREEN}   Example: ${YELLOW}./publish_docs.sh 2.2 ${GREEN}- Prepare docs for 2.2 release.\n"
```
johnlcox: Why is this hardcoded for 2.2?

.travis/publish_docs.sh
```diff
@@ -67,13 +80,14 @@ else
 fi
 
 printf "${CYAN}Generating site.$RESET \n"
-mvn clean site
+mvn clean site > /dev/null
```
johnlcox: Why are we redirecting the output? What if there is an error and we need to troubleshoot?

RELEASING.md
```diff
@@ -2,30 +2,50 @@
 
 ## Preparing for the Release
```
johnlcox: Are all of these steps necessary if I wanted to do a release, or are these steps only necessary for setting up travis once?

jaxrs/pom.xml
```diff
@@ -74,10 +74,6 @@
         </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-api</artifactId>
```
johnlcox: Why?

pom.xml
```diff
@@ -927,19 +927,8 @@
     <profiles>
         <profile>
             <id>ossrh</id>
-            <properties>
```
johnlcox: Is it better to have these things in the settings.xml file instead of this profile?

.travis/publish_docs.sh
sparuvu: otherwise this will fill the logs and will fail the build. It happened several times when I was trying different things

.travis/publish_docs.sh
b-boogaard: it is just used as an example and currently I don't think this file is filtered by maven which makes it harder to inject the latest value. 

with that said we could change this so that we don't use any specific version in the usage printout, or we could use some shell magic to grab the latest from the pom

RELEASING.md
sparuvu: setting it up for the first time. Since this is already set up ...we should really refer to the releasing section.

RELEASING.md
sparuvu: I wasn't very sure about. If you are sure then I can remove this part.

jaxrs/pom.xml
sparuvu: we duplicated this dependency in the pom.

pom.xml
sparuvu: Leaving them in the profile makes the IDE complaining about unset environment variables even though that really doesn't cause any compile/build problems. Its just annoying. So I moved them to settings.xml.

.travis.yml
```diff
@@ -1,15 +1,35 @@
...
+- mvn --settings .travis/settings.xml install -DskipTests=true -Dmaven.javadoc.skip=true -Dgpg.skip=true -B -V -e
+before_install:
```
sparuvu: moved it.

.travis/publish.sh
sparuvu: updated the comment as suggested.

.travis.yml
sparuvu: we need this. I have added it. I only commented certain parts to reduce the build time.

RELEASING.md
johnlcox: I'm fairly certain it won't be an actual requirement for sonatype.

.travis/publish.sh
johnlcox: I don't think it's a problem. This really pushes the tag to a staging area on sonatype and we have to actually go approve the staged artifacts for them to go to maven central.

.travis/publish_docs.sh
johnlcox: My opinion is not strong either. Does anyone else care?

.travis/publish_docs.sh
johnlcox: Ah yes. I should pay more attention when I'm reading.

.travis/publish_docs.sh
johnlcox: Why does it fail? Takes too long?

RELEASING.md
johnlcox: Maybe add a note about that.

.travis/publish_docs.sh
sparuvu: calling it `compile_docs.sh` makes it even better IMO, as we are only prepping the docs to be published. As Github publishes these docs automatically from gh-pages branch it really doesn't make sense to call it publish.

.travis.yml
sparuvu: deleted

RELEASING.md
sparuvu: rephrased the sentence

RELEASING.md
sparuvu: updated

RELEASING.md
sparuvu: updated

.travis/publish_docs.sh
```diff
@@ -88,13 +102,21 @@ cp -r target/staging/* "$release_tag"/
 
 printf "${CYAN}Creating redirecting index.html.$RESET \n"
 mv target/build_index.sh "."
-sh ./build_index.sh "$release_tag"
+bash ./build_index.sh "$release_tag"
```
sparuvu: travis doesn't load `bash` by default and it runs with the minimal version - `dash`. So, executing other scripts with `sh` was causing problem. So had to execute the scripts exclusively with `bash`

johnlcox: This explanation doesn't make sense to me. `sh` should be even more primitive than `bash`.

b-boogaard: Travis CI uses `sh` in one of their examples[1] when talking about custom scripts.

[1] https://docs.travis-ci.com/user/customizing-the-build#Customizing-the-Installation-Step

sparuvu: The problem is not with the script itself. Its how it is being executed if I am understanding it correctly. To be explicit with the problem - The shell that travis loads doesn't have permission to execute this script. Why ? I don't know. Its definitely not with the permissions on the file and I don't want to run the build with sudo. 

Here is the [failed build|https://travis-ci.org/cerner/beadledom/builds/193790722] (open the `deploying application` stage in the log to see the error). Here is the immediate [build that passed|https://travis-ci.org/cerner/beadledom/builds/193797533]. Here is the diff between the two builds - https://github.com/cerner/beadledom/compare/7e293aaadf68...7cefc4f57d35

sparuvu: I tried updating the shebang to use `#!/usr/bin/env sh` as suggested offline and tried running it. It gave me this error - `/home/travis/build/cerner/beadledom/.travis//publish_docs.sh: 22: /home/travis/build/cerner/beadledom/.travis//publish_docs.sh: Syntax error: "(" unexpected
`

Note: I tried on the different branch. Here is the diff - https://github.com/cerner/beadledom/compare/89d1955ea347...8251cf382e9b

PR #11, opened 2017-01-27
Fix the Sonatype validation issues.

sparuvu:
# What? Why?
Due to some configuration errors, the javadocs weren't being pushed and also the artifacts were not being signed. This PR contains the fix to these issues.

# How was it tested?
released a beta version. 

Note: did not sync'ed the artifact to maven central.

 Reviewers: 

Reviewers
----


.travis.yml
```diff
@@ -8,18 +8,18 @@ before_install:
 - export PATH=$PATH:${TRAVIS_BUILD_DIR}/target
 install:
 - sudo apt-get install -y gnupg2
-- mvn --settings .travis/settings.xml install -DskipTests=true -Dmaven.javadoc.skip=true -Dgpg.skip=true -B -V -e
-before_deploy:
-- openssl aes-256-cbc -K $encrypted_4d3aca009c62_key -iv $encrypted_4d3aca009c62_iv -in $GPG_DIR/deploy_site.enc -out $GPG_DIR/deploy_site -d
-- openssl aes-256-cbc -pass pass:$ENCRYPTION_PASSWORD -in $GPG_DIR/pubring.gpg.enc -out $GPG_DIR/pubring.gpg -d
-- openssl aes-256-cbc -pass pass:$ENCRYPTION_PASSWORD -in $GPG_DIR/secring.gpg.enc -out $GPG_DIR/secring.gpg -d
+script:
+- mvn install -DskipTests=true -Dmaven.javadoc.skip=true -Dgpg.skip=true -B -V -e
```
johnlcox: I don't think `gpg.skip=true` is needed since the profile with gpg shouldn't be activated here.

johnlcox: We shouldn't skip tests here.

.travis/publish.sh
```diff
@@ -1,11 +1,22 @@
...
+    openssl aes-256-cbc -pass pass:$ENCRYPTION_PASSWORD -in $GPG_DIR/pubring.gpg.enc -out $GPG_DIR/pubring.gpg -d
+    openssl aes-256-cbc -pass pass:$ENCRYPTION_PASSWORD -in $GPG_DIR/secring.gpg.enc -out $GPG_DIR/secring.gpg -d;
+
+    echo "git checking out $TRAVIS_TAG"
     git checkout tags/$TRAVIS_TAG -b tag_$TRAVIS_TAG
-    mvn deploy --settings $GPG_DIR/settings.xml -DskipTests=true -B -U
-    ${GPG_DIR}/publish_docs.sh $TRAVIS_TAG
+
+    echo "deploying $TRAVIS_TAG to maven central"
+    mvn deploy --settings $GPG_DIR/settings.xml -DskipTests=true -DattachScaladocs=true -B -U
```
johnlcox: It may be okay to skip tests here if we are running them above, but does it really hurt to run them again?

.travis/publish.sh
```diff
@@ -1,11 +1,22 @@
...
 if [ ! -z "$TRAVIS_TAG" ]
 then
-    echo "deploying beadledom $TRAVIS_TAG to maven central"
+    echo "Decrypting secrets ..."
+    openssl aes-256-cbc -K $encrypted_4d3aca009c62_key -iv $encrypted_4d3aca009c62_iv -in $GPG_DIR/deploy_site_key.enc -out $GPG_DIR/deploy_site_key -d
+    openssl aes-256-cbc -pass pass:$ENCRYPTION_PASSWORD -in $GPG_DIR/pubring.gpg.enc -out $GPG_DIR/pubring.gpg -d
+    openssl aes-256-cbc -pass pass:$ENCRYPTION_PASSWORD -in $GPG_DIR/secring.gpg.enc -out $GPG_DIR/secring.gpg -d;
+
+    echo "git checking out $TRAVIS_TAG"
     git checkout tags/$TRAVIS_TAG -b tag_$TRAVIS_TAG
```
johnlcox: Wouldn't it already be on the tag if the `TRAVIS_TAG` variable is set?

.travis.yml
sparuvu: sure removed it here - https://github.com/cerner/beadledom/pull/11/commits/92fbbbd239c5ac466b25de461c99592b934ce941. I think I was skipping them to speed up the builds. I agree we shouldn't skip them

johnlcox: What about gpg skip?

.travis/publish.sh
sparuvu: technically speaking travis doesn't  checkout the tag. It checksout the master and switches to a specific commit at which the tag was cut.  I guess they both resolve to the same command internally. Removed checking  out tag explicitly.

.travis/publish.sh
sparuvu: we can run the tests again. Removed the skip tests option from the command.

.travis.yml
sparuvu: removed it in the same commit as above.

Issue #12, opened 2017-01-27
ModuleProcess of Injector infinite loop


johnlcox:
The initialization of Resteasy resources using the `ModuleProcessor` in `ResteasyContextListener` could cause an infinite loop if the Guice Injector has parent injectors.

The fix is to reassign the parent injector back to the `injector` variable instead of creating a `parent` variable. Testing is the harder bit. Before fixing this we should identify a way to test the current behavior and write a failing test.

See https://github.com/cerner/beadledom/blob/master/resteasy/src/main/java/com/cerner/beadledom/resteasy/ResteasyContextListener.java#L51-L55

johnlcox:
Fixed by #13

PR #13, opened 2017-01-28
Fix infinite loop in ResteasyContextListener when injector has parent

jpeg:

An infinite loop would occur in `ResteasyContextListener` if the Guice injector had a parent injector.

How was it tested?
----

Added new unit test and verified that it resulted in an infinite loop before the fix was implemented.

How to test
----

- [ ] `mvn clean install -U`

Reviewers
----

### DIFF
--- a/resteasy/src/main/java/com/cerner/beadledom/resteasy/ResteasyContextListener.java
+++ b/resteasy/src/main/java/com/cerner/beadledom/resteasy/ResteasyContextListener.java
@@ -5,6 +5,8 @@ import com.cerner.beadledom.configuration.ConfigurationSource;
 import com.cerner.beadledom.lifecycle.GuiceLifecycleContainers;
 import com.cerner.beadledom.lifecycle.LifecycleContainer;
 import com.cerner.beadledom.lifecycle.LifecycleInjector;
+
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
 import com.google.inject.Injector;
 import com.google.inject.Module;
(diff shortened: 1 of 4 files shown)


resteasy/src/main/java/com/cerner/beadledom/resteasy/ResteasyContextListener.java
```diff
@@ -76,4 +70,21 @@ protected void withInjector(Injector injector) {
    * Returns the list of Guice modules to be used with Resteasy.
    */
   protected abstract List<? extends Module> getModules(ServletContext context);
+
+  /**
+   * Processes module bindings in the Guice injector.
+   */
+  @VisibleForTesting
+  public void processInjector(ServletContext context, Injector injector) {
```
sparuvu: somehow I am not comfortable having this method. Since this is the initial point of a Web application can the Injector that is created in this class even have a parent Injector ? What if we remove this loop completely  and just have the processor process the injector that is created in this class ?

sparuvu: The other way to detect a loop is to use  run a fast-slow runner technique to detect the loop in a LinkedList. Seems like a overkill but may be worth trying it.

johnlcox: I'm not convinced the infinite loop could happen at all or that the while loop is even needed at all. Since the context listener creates the injector is there even any possible way for the injector to have a parent injector?

nimeshsubramanian:
issue #12

resteasy/src/main/java/com/cerner/beadledom/resteasy/ResteasyContextListener.java
jpeg: Removed the while loop: https://github.com/cerner/beadledom/pull/13/commits/b666e0add8a35878245812f04789793a83a02fc6

PR #16, opened 2017-01-31
WebApplicationExceptionMapper is called when exceptions are thrown by the checkParam method

nimeshsubramanian:

The WebApplicationExceptionMapper is only called when there isnt a message set in the response.

relevant code from ExceptionHandler.java
```
      if (unwrappedException instanceof WebApplicationException) {
         WebApplicationException wae = (WebApplicationException) unwrappedException;
         if (wae.getResponse() != null && wae.getResponse().getEntity() != null) return wae.getResponse();
      }

      if ((jaxrsResponse = executeExceptionMapper(unwrappedException)) != null) {
         return jaxrsResponse;
      }
```

So with `checkParam` setting the messages, our exception mapper's are not being called. So am removing the message from the response and using a constructor of WebApplicationException that let's me pass both the message and a response.

How was it tested?
----

tested on a local service.

How to test
----

- [ ] `mvn clean install -U`

Reviewers
----

- [x] [Supriya Lal](https://github.com/lal-s)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Beadledom Changelog
 
-## 2.4
+## 2.4 - In Development
 
 ### Defects Corrected
 * WebApplicationException mapper is called when exceptions are thrown by the checkParam method in JaxRsParamConditions class.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Beadledom Changelog
 
+## 2.4
+
+### Defects Corrected
+* WebApplicationException mapper is called when exceptions are thrown by the checkParam method in JaxRsParamConditions class.
+
 ## 2.3 - 24 01 2017
 
 ### Enhancements
--- a/jaxrs/src/main/java/com/cerner/beadledom/jaxrs/JaxRsParamConditions.java
+++ b/jaxrs/src/main/java/com/cerner/beadledom/jaxrs/JaxRsParamConditions.java
@@ -27,6 +27,7 @@ import javax.ws.rs.core.Response;
  * </pre>
  *
  * @author John Leacox
+ * @author Nimesh Subramanian
  */
 public class JaxRsParamConditions {
   private JaxRsParamConditions() {
--- a/jaxrs/src/main/java/com/cerner/beadledom/jaxrs/JaxRsParamConditions.java
+++ b/jaxrs/src/main/java/com/cerner/beadledom/jaxrs/JaxRsParamConditions.java
@@ -43,10 +43,9 @@ public class JaxRsParamConditions {
   public static void checkParam(boolean expression, Object errorMessage) {
     if (!expression) {
       Response response = Response.status(Response.Status.BAD_REQUEST)
-          .entity(String.valueOf(errorMessage))
           .type(MediaType.TEXT_PLAIN)
           .build();
-      throw new WebApplicationException(response);
+      throw new WebApplicationException(String.valueOf(errorMessage), response);
     }
   }
 }
--- a/jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/JaxRsParamConditionsSpec.scala
+++ b/jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/JaxRsParamConditionsSpec.scala
@@ -2,23 +2,25 @@ package com.cerner.beadledom.jaxrs
 
 import javax.ws.rs.WebApplicationException
 import javax.ws.rs.core.MediaType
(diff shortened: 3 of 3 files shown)


jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/JaxRsParamConditionsSpec.scala
```diff
@@ -17,7 +17,6 @@ class JaxRsParamConditionsSpec extends FunSpec with ShouldMatchers {
           JaxRsParamConditions.checkParam(false, "some message")
         }
         exception.getResponse.getStatus should be(400)
-        exception.getResponse.getEntity should be("some message")
```
jpeg: Should check the exception message instead.

nimeshsubramanian: added: 9976847

lal-s: +1

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Beadledom Changelog
 
+## 2.4
```
johnlcox: Maybe add `- In Development`

nimeshsubramanian: updated: ab4ee86

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/JaxRsParamConditions.java
```diff
@@ -43,10 +43,9 @@ private JaxRsParamConditions() {
   public static void checkParam(boolean expression, Object errorMessage) {
```
sparuvu: I am curious why does this have to be an `Object`. Why not String ? It seems like we are converting explicitly to string.

sparuvu: update the @author section as well.

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/JaxRsParamConditionsSpec.scala
```diff
@@ -17,8 +17,8 @@ class JaxRsParamConditionsSpec extends FunSpec with ShouldMatchers {
...
         exception.getResponse.getMediaType should be(MediaType.TEXT_PLAIN_TYPE)
+        exception.getMessage should be("some message")
```
sparuvu: can you please update this spec to use MustMatchers instead of should ?

Edit: update the it blocks  description as well.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/JaxRsParamConditions.java
nimeshsubramanian: Im not sure why it's an Object instead of a `String`.  Changing that now, would make it a non-passive change. 

Should this method be added to the docs?

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/JaxRsParamConditionsSpec.scala
nimeshsubramanian: updated the matcher: 201987b

I think that the description in the it block is still valid.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/JaxRsParamConditions.java
nimeshsubramanian: added the author tag: 201987b

johnlcox: It's `Object` because this is similar to the methods in the Guava class `Preconditions` so I stuck with a similar signature.

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/JaxRsParamConditionsSpec.scala
sparuvu: I mean it should be something like `it must be ... ` instead of `it should be`

nimeshsubramanian: changed: https://github.com/cerner/beadledom/pull/16/commits/64f16bd779e03367c14f9d11b25c2d3d2109594a

PR #19, opened 2017-02-07
Adding SCM Version in health check page

rajkumarts:
https://github.com/cerner/beadledom/issues/20

* Added the SCM Version to health check. In development environment many developers deploy their own version of services. This field will be helpful in identifying which scm version of the service is deployed. 

How was it tested?
----

* Unit tests. Also consumed this snapshot version in our project to make sure the correct scm version is displayed. 

How to test
----

* Consume this version in any of the services that uses beadledom health and check the 
`{service-url}/meta/health/diagnostic`.

- [ ] `mvn clean install -U`

Reviewers
----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Beadledom Changelog
 
 ## 2.4 - In Development
+* Added SCM Revision to health check.
 
 ### Defects Corrected
 * Removed unnecessary loop in `ResteasyContextListener` which would cause infinite loop if run.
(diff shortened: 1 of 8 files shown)


b-boogaard:
I'll need to double check, but I think something like this might need to go into a `meta` object.

ntlanglois:
+1

sparuvu:
It would be good if we can add support to Beadledom to accept any arbitrary fields via a Guice Multibinders so the consumers can set whatever properties they need as part of their health checks. Since this change is not going to be trivial and need some refactoring to the code in this branch,  I am gonna close this. Please open a new PR with the sugesstion. Please free to open another PR with the new design.

PR #22, opened 2017-02-09
Fixed BeadledomClientBuilderProvider using wrong injector

johnlcox:

`BeadledomClientBuilderProvider` was using a `Provider<Injector>` to get an instance of a Guice `Injector`. Since the provider defers getting an instance of an `Injector` until the provider is actually used, the context private module child Injectors is lost by the time an instance is retrieved. This provider should be using the nearest child injector from the private modules the client is being configured in. Changing the class to take an instance of `Injector` directly instead of a provider directly makes the injection more eager and results in the correct child injector being used.

How was it tested?
----

Additional unit test.

How to test
----

- [ ] `mvn clean install -U`

Reviewers
----

### DIFF
--- a/client/beadledom-client-guice/src/main/java/com/cerner/beadledom/client/BeadledomClientBuilderProvider.java
+++ b/client/beadledom-client-guice/src/main/java/com/cerner/beadledom/client/BeadledomClientBuilderProvider.java
@@ -36,7 +36,7 @@ class BeadledomClientBuilderProvider implements Provider<BeadledomClientBuilder>
       BeadledomClientBuilder.class,
       BeadledomClientLifecycleHook.class);
 
-  private Provider<Injector> injectorProvider;
+  private Injector injector;
   private String correlationIdHeader;
 
   BeadledomClientBuilderProvider(Class<? extends Annotation> clientBindingAnnotation) {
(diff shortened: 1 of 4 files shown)


client/beadledom-client-guice/src/main/java/com/cerner/beadledom/client/BeadledomClientBuilderProvider.java
```diff
@@ -84,15 +84,19 @@ public BeadledomClientBuilder get() {
       }
     }
 
-    Injector injector = injectorProvider.get();
-    processInjector(injector, clientBuilder);
-    while (injector.getParent() != null) {
-      injector = injector.getParent();
-      processInjector(injector, clientBuilder);
+    Injector tempInjector = getInjector();
```
sparuvu: why can't we just use the private variable directly instead of a getter ?

johnlcox: I put it in a getter to avoid an issue of accidentally modifying the injector field in the while loop.

Ideally the injector field would be `final`, but cannot be since we must use method injection to initialize this class. Using a getter seemed like a decent alternative to avoid accidental modification.

sparuvu:
+1

b-boogaard:
+1

PR #27, opened 2017-03-30
Cleaned up dependencies and fixed few minor issues with generated cod…

bbaugher:
…e in archetype

 * Removed unused dependencies and added a few dependencies that weren't declared but used
 * The HelloWorldDto wouldn't work with Jackson as is since theres no arg-less constructor so had to had the `@JsonCreator` for the create method
 * The versions for slf4j were specified only in service which is inconsistent with everything else. You also don't need log4j just the slf4j-log4j12
 * The `ResteasyBootstrapModule` was called out in findbugs for [usage of getClass with getResourceAsStream](http://findbugs.sourceforge.net/bugDescriptions.html#UI_INHERITANCE_UNSAFE_GETRESOURCE) and that this could behave differently if the class was extended. This was fixed by referencing the class itself instead of using `getClass()`.

I also attempted to update the changelog but I'm not totally sure I got the format right

How was it tested?
----

Tested this by creating a new project, making the changes, started the service using `mvn tomcat7:run-war` and then running a test using the client to interact with the service

How to test
----

- [ ] `mvn clean install -U`

Reviewers
----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Beadledom Changelog
 
+## 2.5 - ???
+
+### Defects Corrected
+ * Cleaned up dependencies and fixed few minor issues with generated code in archetype 
+
 ## 2.4 - 16 Feb 2017
 
 ### Additions
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -11,6 +11,7 @@
 * Supriya Lal [@lal-s][supriya-lal]
 * Jason Gassel [@jpeg][jason-gassel]
 * Andy Nelson [@anelson425][andy-nelson]
+* Bryan Baugher [@bbaugher][bryan-baugher]
 
 [john-leacox]: https://github.com/johnlcox
 [jacob-williams]: https://github.com/brokensandals
@@ -23,3 +24,4 @@
 [supriya-lal]: https://github.com/lal-s
 [jason-gassel]: https://github.com/jpeg
 [andy-nelson]: https://githum.com/anelson425
+[bryan-baugher]: https://githum.com/bbaugher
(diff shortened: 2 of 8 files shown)


CONTRIBUTORS.md
```diff
@@ -23,3 +24,4 @@
 [supriya-lal]: https://github.com/lal-s
 [jason-gassel]: https://github.com/jpeg
 [andy-nelson]: https://githum.com/anelson425
+[bryan-baugher]: https://githum.com/bbaugher
```
johnlcox: Looks like Andy put the wrong URL `githum` and it got copied for yours as well.

johnlcox:
+1

bbaugher:
Fixed the contributor url

sparuvu:
+1

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Beadledom Changelog
 
+## 2.5 - ???
```
nimeshsubramanian: NIT: should it be `In Developement` rather than `???`

nimeshsubramanian:
+1

archetype/simple-service/src/main/resources/archetype-resources/api/src/main/java/api/model/HelloWorldDto.java
```diff
@@ -27,7 +29,8 @@ private HelloWorldDto(String name, String message) {
    * @param message the message to be returned
    * @return {@link HelloWorldDto} encapsulating a name and message
    */
-  public static HelloWorldDto create(String name, String message) {
+  @JsonCreator
```
b-boogaard: Good catch here.

b-boogaard:
+1

PR #28, opened 2017-03-30
Added toString implementation to DelegatingGenericResponse

bbaugher:

Dumb but nice to have, added toString implementation to help testing and debugging. 

I originally did this while playing around with the client and implemented it on what the client sees (DelegatingGenericResponse) but I'm thinking this should instead be implemented on the interface (GenericResponse) instead otherwise you would likely want to implement toString on BuiltGenericResponse to be consistent. Let me know if you have a preference.

I also effectively implemented toString for raw Response otherwise you just get a reference to ApacheHttpClient4Dot3Engine on the client side and from what I can tell BuiltResponse on server side. I can't decide if I should try to provide a pull to resteasy to implement toString for those classes or just leave it here. Again let me know if you have a preference.

How was it tested?
----

Built locally and ran with a client implementation

How to test
----

- [ ] `mvn clean install -U`

Reviewers
----

### DIFF
--- a/jaxrs-genericresponse/src/main/java/com/cerner/beadledom/jaxrs/DelegatingGenericResponse.java
+++ b/jaxrs-genericresponse/src/main/java/com/cerner/beadledom/jaxrs/DelegatingGenericResponse.java
@@ -25,6 +25,7 @@ public class DelegatingGenericResponse<T> extends Response implements GenericRes
   private final T body;
   private final Class<?> bodyClass;
   private final ErrorBody errorBody;
+
   private final Response rawResponse;
 
   /**
@@ -249,5 +250,16 @@ public class DelegatingGenericResponse<T> extends Response implements GenericRes
   @Override
   public String getHeaderString(String name) {
     return rawResponse.getHeaderString(name);
+  }
+
+  @Override
+  public String toString() {
+    return "DelegatingGenericResponse{body=" + body + ", bodyClass=" + bodyClass + ", errorBody=" + errorBody
+            + ", rawResponse=Response{status=" + rawResponse.getStatus() + ", mediaType="
+            + rawResponse.getMediaType() + ", date=" + rawResponse.getDate() + ", length=" + rawResponse.getLength()
+            + ", lastModified=" + rawResponse.getLastModified() + ", entityTag=" + rawResponse.getEntityTag()
+            + ", language=" + rawResponse.getLanguage() + ", location=" + rawResponse.getLocation()
+            + ", headers=" + rawResponse.getHeaders() + ", cookies=" + rawResponse.getCookies() + ", links="
+            + rawResponse.getLinks() + " } }";
   }
 }
--- a/jaxrs-genericresponse/src/main/java/com/cerner/beadledom/jaxrs/DelegatingGenericResponse.java
+++ b/jaxrs-genericresponse/src/main/java/com/cerner/beadledom/jaxrs/DelegatingGenericResponse.java
@@ -254,6 +254,6 @@ public class DelegatingGenericResponse<T> extends Response implements GenericRes
   @Override
   public String toString() {
     return "DelegatingGenericResponse{body=" + body + ", bodyClass=" + bodyClass + ", errorBody=" + errorBody
-            + ", rawResponse=" + ResponseToStringWrapper.toString(rawResponse) + " }";
+            + ", rawResponse=" + new ResponseToStringWrapper(rawResponse) + " }";
   }
 }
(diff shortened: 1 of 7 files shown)


jaxrs-genericresponse/src/main/java/com/cerner/beadledom/jaxrs/DelegatingGenericResponse.java
```diff
@@ -25,6 +25,7 @@
   private final T body;
   private final Class<?> bodyClass;
   private final ErrorBody errorBody;
+
```
sparuvu: can you please remove this unnecessary extra empty line.

sparuvu:
can you please update the changelog with this change ?

sparuvu:
`JAXRS-genericResponse` is in Java 6 so we can't have the default methods on the interface :(. I think, Implementing`toString` on the `DelegatingGenericResponse` is fine I guess. It would be great if we could add `toString` on `BuiltGenericResponse` as well.

jaxrs-genericresponse/src/main/java/com/cerner/beadledom/jaxrs/DelegatingGenericResponse.java
bbaugher: Done, https://github.com/cerner/beadledom/pull/28/commits/08ddce40ddfc37c67b994dbde0f54f91d014f62b

bbaugher:
Added to changelog, https://github.com/cerner/beadledom/pull/28/commits/115608019cabb55acb48685d31eceda510035c37

Implemented toString in BuiltGenericResponse, https://github.com/cerner/beadledom/pull/28/commits/159a76e1d482a8ebe12cc253aedbaccc74880ebf

nimeshsubramanian:
+1

sparuvu:
I am wondering if we should add a test for each class 🤔

b-boogaard:
+1

resteasy-genericresponse/src/main/java/com/cerner/beadledom/resteasy/BuiltGenericResponse.java
```diff
@@ -255,4 +256,15 @@ public Link getLink(String relation) {
   public String getHeaderString(String name) {
     return rawResponse.getHeaderString(name);
   }
+
+  @Override
+  public String toString() {
+    return "BuiltGenericResponse{body=" + body + ", bodyClass=" + bodyClass + ", errorBody=" + errorBody
+            + ", rawResponse=Response{status=" + rawResponse.getStatus() + ", mediaType="
```
johnlcox: Rather than duplicating the `rawResponse` String code, it might be better to have a class like `ResponseToStringWrapper` that only has a `toString` method on it.

bbaugher: Done, https://github.com/cerner/beadledom/pull/28/commits/3a04e401c758fe3f8e1b39405bf0581995634f41

bbaugher:
@sparuvu If you would like, because theres almost no logic I had left it out

jaxrs-genericresponse/src/main/java/com/cerner/beadledom/jaxrs/ResponseToStringWrapper.java
```diff
@@ -0,0 +1,26 @@
+package com.cerner.beadledom.jaxrs;
+
+import javax.ws.rs.core.Response;
+
+/**
+ * Simple class to implement {@link Response} to String.
+ */
+public class ResponseToStringWrapper {
```
johnlcox: This is more of a utility class than a wrapper. I was imagining a class that takes in `Response` to it's constructor and then implements `Object#toString()`.

sparuvu:
> @sparuvu If you would like, because theres almost no logic I had left it out

I prefer to have this code covered.

bbaugher:
Added tests and converted the wrapper to implement toString and take in Response in the constructor, let me know if theres anything else

resteasy-genericresponse/src/test/scala/com/cerner/beadledom/resteasy/BuiltGenericResponseSpec.scala
```diff
@@ -436,5 +440,30 @@ class BuiltGenericResponseSpec
         verify(rawResponse).getHeaderString("header")
       }
     }
+
+    describe("#toString") {
+      it("returns String with all fields included") {
+        val rawResponse = mock[Response]
+        when(rawResponse.getStatus).thenReturn(200)
+        when(rawResponse.getMediaType).thenReturn(MediaType.APPLICATION_JSON_TYPE)
+        when(rawResponse.getDate).thenReturn(new Date(0L))
```
johnlcox: Looks like the dates are failing on Travis due to timezone differences. Might be best to just get the expected dates programmatically as well.

bbaugher: Should be fixed now, https://github.com/cerner/beadledom/pull/28/commits/bd07cae96b21c4610e955327b3f221aebbe23bd8

johnlcox:
+1

PR #33, opened 2017-05-04
Fix dependency version issue in archetype

lal-s:

Add the dependency version for guava and javax to the parent pom of the archetype

How was it tested?
----

Generate a service through archetype and running it successfully

How to test
----

- [ ] `mvn clean install -U`

Reviewers
----

### DIFF
--- a/archetype/simple-service/src/main/resources/archetype-resources/pom.xml
+++ b/archetype/simple-service/src/main/resources/archetype-resources/pom.xml
@@ -82,6 +82,11 @@
                 <version>2.0.2</version>
             </dependency>
             <dependency>
+                <groupId>com.google.guava</groupId>
+                <artifactId>guava</artifactId>
+                <version>15.0</version>
+            </dependency>
+            <dependency>
                 <groupId>com.google.inject</groupId>
                 <artifactId>guice</artifactId>
                 <version>${guice.version}</version>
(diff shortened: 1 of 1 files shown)


archetype/simple-service/src/main/resources/archetype-resources/pom.xml
```diff
@@ -82,6 +83,11 @@
                 <version>2.0.2</version>
             </dependency>
             <dependency>
+                <groupId>com.google.guava</groupId>
+                <artifactId>guava</artifactId>
+                <version>${guava.version}</version>
```
johnlcox: Since there is only one guava jar the version can be embedded here rather than using a property.

The value of the property is for cases like guice where there are extension jars with matching versions (e.g. guice-multibindings).

lal-s: [4715e94](https://github.com/cerner/beadledom/commit/4715e94705ce39e6ca4f902abf1cb1346b7f65b5)

johnlcox:
+1

archetype/simple-service/src/main/resources/archetype-resources/pom.xml
```diff
@@ -82,6 +82,11 @@
                 <version>2.0.2</version>
             </dependency>
             <dependency>
+                <groupId>com.google.guava</groupId>
+                <artifactId>guava</artifactId>
+                <version>15.0</version>
```
b-boogaard: Is there anything keeping us from going higher with this? I think `21` is the latest.

johnlcox: This matches the version in the beadledom parent pom. Should we up that version too?

Currently beadledom-client and libraries shared between client and server are Java 6 compatible, so if we did bump the guava version we can only go as far as the lateast one supporting Java 6.

lal-s: According to the offline discussion, we are sticking to version 15

Issue #21, opened 2017-02-08
Add JAX-RS PATCH annotation


johnlcox:
JAX-RS does not have a PATCH annotation for supporting HTTP patch method like it does for others like GET or POST. We should add a PATCH annotation that can be used for handling the patch HTTP method.

b-boogaard:
+1 to this.

nimeshsubramanian:
PR: https://github.com/cerner/beadledom/pull/36

b-boogaard:
Merge to master: e83ff7ce05c6e0ddd9d9d1fe17b3c3a04a200cfc

PR #36, opened 2017-05-15
Add JAX-RS PATCH annotation

EricChristensen:
From [issue 21](https://github.com/cerner/beadledom/issues/21) JAX-RS does not have a PATCH annotation for supporting HTTP patch method like it does for others like GET or POST. We should add a PATCH annotation that can be used for handling the patch HTTP method.

How was it tested?
----
The PATCH annotation was tested by consuming this change in a test branch of a project that already consumes beadledom. In this project, a simple resource that uses the PATCH annotation was used to manually make sure that the behavior performed as expected. Two unit tests were also added to the project to 1. Ensure that PATCH is used and behaves as expected and 2. PATCH is used and performs an operation on the server to ensure that serialization works as expected.

How to test
----
- [ ] `mvn clean install -U`

Reviewers
----

- [x] [Brian van de Boogaard](https://github.com/b-boogaard)


jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/PATCHSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+
+import scala.collection.JavaConverters._
+import javax.ws.rs.core.MediaType
+
+/**
+  * Spec tests for {@link PATCH}.
+  *
+  * @author Eric Christensen
+  */
+class PATCHSpec extends FunSpec with BeforeAndAfterAll with ShouldMatchers with MockitoSugar {
```
johnlcox: I'd like to see a test with an actual stood up service. Might fit in https://github.com/cerner/beadledom/blob/c2125c7228f9a778b56a07615f00a4d729a86270/client/beadledom-client-test/src/test/scala/com/cerner/beadledom/client/ClientServiceSpec.scala

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/PATCHSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+        request.contentType(MediaType.APPLICATION_JSON)
+        request.content(new ByteArrayInputStream(mapper.writeValueAsBytes(model)))
+
+        val response: MockHttpResponse = new MockHttpResponse()
+        dispatcher.invoke(request, response)
+
+        response.getStatus shouldBe 200
+      }
+
+      it("should change the request model fields") {
```
kyleroush: Why do we need to have these two test, when you are calling the function with the same input twice?

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/PATCHSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+        request.contentType(MediaType.APPLICATION_JSON)
+        request.content(new ByteArrayInputStream(mapper.writeValueAsBytes(model)))
+
+        val response: MockHttpResponse = new MockHttpResponse()
+        dispatcher.invoke(request, response)
+
+        response.getStatus shouldBe 200
+
+        val newModel: FakeModel = mapper.readValue(response.getContentAsString, classOf[FakeModel])
+        newModel.name shouldBe "newName"
```
kyleroush: In theory, if this was a real patch shouldn't these responses be what you sent to the patch?

jaxrs/src/test/java/com/cerner/beadledom/jaxrs/provider/FakeModel.java
```diff
@@ -28,6 +28,10 @@
   @JsonProperty("inner_models")
   public List<FakeInnerModel> innerModels;
 
+  public FakeModel() {
```
kyleroush: why did you add a default constructor? the only place I see you using it is here where you use the constructor that already exist.

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/PATCHSpec.scala
EricChristensen: The PATCH resource updates the request that was sent to it. Normally the update would be to an existing resource in a database, but for the purposes of just performing an operation on the server this is fine.

jaxrs/src/test/java/com/cerner/beadledom/jaxrs/provider/FakeModel.java
EricChristensen: The default constructer and the setters were added to make the creating of the FakeModel more like the builder pattern. It is used in the Spec at the recommendation of Brian.

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/PATCHSpec.scala
EricChristensen: They are indeed the same function calls with the same parameters but it seemed appropriate to have two different "it" descriptors, one for just making sure that PATCH is called successfully and one to make sure that serialization takes place properly by making sure the server performed a change and returned it.

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/PATCHSpec.scala
EricChristensen: PATCH used with an actual stood up service in the "retrieves the resources from different clients test" [d582c3](https://github.com/cerner/beadledom/pull/36/commits/d582c3a78cc5f6ca53ed2ee53d7c9507bdb69565)

jaxrs/src/test/java/com/cerner/beadledom/jaxrs/provider/FakeResource.java
```diff
@@ -0,0 +1,24 @@
...
+  @PATCH
+  @Path("/Patch")
+  @Produces(MediaType.APPLICATION_JSON)
+  @Consumes(MediaType.APPLICATION_JSON)
+  public Response fakePatch(FakeModel model) {
+    model.setId("newId");
+    model.setName("newName");
+    return Response.ok(model).build();
+  }
+
```
b-boogaard: nit: extra newline

jaxrs/src/test/java/com/cerner/beadledom/jaxrs/provider/FakeModel.java
b-boogaard: We didn't break of the existing tests using this model did we? I can't imagine so but you never know.

jaxrs/src/test/java/com/cerner/beadledom/jaxrs/provider/FakeResource.java
EricChristensen: [48dafb3](https://github.com/cerner/beadledom/pull/36/commits/48dafb39c647b45e9fee3f4e749bf787ab9b2711)

jaxrs/src/test/java/com/cerner/beadledom/jaxrs/provider/FakeModel.java
EricChristensen: Nope all tests pass even after model change

b-boogaard:
I'm not sure exactly what we would say about it, but it seems like we would want to add the fact we added a `@PATCH` annotation to the [manual for the jax-rs module](http://engineering.cerner.com/beadledom/2.4/docs/manual/jaxrs.html)

EricChristensen:
@b-boogaard what do you think of this? [a1fac9f](https://github.com/cerner/beadledom/pull/36/commits/a1fac9f9a46f0a4ffe59c34ce5ec9c528c42ebb5)

b-boogaard:
I think I would add it either above or below the `Correlation-Id` header. I think I would want to link to https://tools.ietf.org/html/rfc5789 the rfc that added PATCH to HTTP. Then give an interface like example of a resource using it, pointing out it can be used just like the others that are defined by JAX-RS.

@johnlcox @sparuvu @lal-s @nimeshsubramanian thoughts?

EricChristensen:
@b-boogaard [Interface example and link to RFC](https://github.com/EricChristensen/beadledom/blob/f8347b058f0fe313c978098ecb60f15d5c2c73bc/docs/source/manual/jaxrs.rst)

b-boogaard:
How about something like:

[JAX-RS 2.0 (see section 3.3 for Resource Methods)](http://download.oracle.com/otn-pub/jcp/jaxrs-2_0_rev_A-mrel-eval-spec/jsr339-jaxrs-2.0-final-spec.pdf?AuthParam=1494975982_179302191fa8833291d2b6647856d11b) does not require implementations to support the `PATCH` HTTP method. This is likely do to the fact that `PATCH` was introduced in a later [rfc](https://tools.ietf.org/html/rfc5789) that added the new HTTP method to the already existing HTTP/1.1 specification.

`@PATCH` was added to [beadledom-jaxrs](https://github.com/cerner/beadledom/tree/master/jaxrs) to allow services to support partial updates without the need of overloading `@POST`. The annotation has no opinion on how the service decides to implement the resource performing the `PATCH` operation. Implementing services have the freedom to support [JSON Patch](https://tools.ietf.org/html/rfc6902) and/or [JSON Merge Patch](https://tools.ietf.org/html/rfc7386).

As long as a service has `beadledom-jaxrs` as a dependency `@PATCH` can be used just like any of the HTTP method annotations defined by JAX-RS. Below is a small example of `@PATCH` being used in an interface for a resource.

`code example here`

I think this could use a bit a clean up but I feel like it reads a bit better. Thoughts?

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/PATCH.java
```diff
@@ -0,0 +1,14 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import javax.ws.rs.HttpMethod;
+
+@Target({ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+@HttpMethod("PATCH")
+public @interface PATCH {
```
sparuvu: could you please add some javadoc to this.

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/PATCHSpec.scala
sparuvu: I would merge the two into one test. We are not testing two different aspects of the method here. Status code 200 means the request is served successfully. For PATCH a successful request mean that the resource is patched correctly. So, having these two checks in one test makes sense IMO.

docs/source/manual/jaxrs.rst
```diff
@@ -26,6 +26,30 @@ To include the correlation id in the catalina.out file using log4j, update the l
...
+  @PATCH
+  @Path("path/to/patch")
+  @Produces(MediaType.APPLICATION_JSON)
+  public Response patch(
+    @PathParam("id") final Long id,
+    @ApiParam(value = "changes to make to the object with the specified id")
+    PatchObject patchObject);
+
+.. _RFC: https://tools.ietf.org/html/rfc5789
+
```
sparuvu: How about this?

```
PATCH
--------------

`PATCH <https://github.com/ktoso/maven-git-commit-id-plugin>`_ is a HTTP verb (similar to GET, POST, DELETE, PUT) to push partial changes to the REST resource. JAX-RS by default doesn't support PATCH hence Beadledom adds the support for PATCH explicitly.

For more details on the PATCH method visit the PATCH RFC_.

To use the PATCH method and annotation for your resource, simply annotate your JAX-RS resource method with the PATCH annotation as shown below

.. code-block:: java

 @PATCH
 @Path("path/to/patch")
 @Produces(MediaType.APPLICATION_JSON)
 public Response patch(
   @PathParam("id") final Long id,
   @ApiParam(value = "changes to make to the object with the specified id")
   PatchObject patchObject);

.. _RFC: https://tools.ietf.org/html/rfc5789
```

EricChristensen:
@sparuvu comments addressed [here](https://github.com/cerner/beadledom/pull/36/commits/7931147398b85634dec536301172c529873eb276). @b-boogaard I went with Sundeeps description of the annotation in the jaxrs.rst. Concise, simple and exactly what the user needs to know. I removed the [git url](https://github.com/ktoso/maven-git-commit-id-plugin/tree/master/src/main/java/pl/project13/maven/git) that was linked in the description because don't see how it relates to PATCH at all...

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/PATCH.java
```diff
@@ -0,0 +1,19 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import javax.ws.rs.HttpMethod;
+
+/**
+ * Indicates that the annotated method responds to HTTP PATCH requests.
+ *
+ * @author John Leacox
```
sparuvu: I think this should be you :) and not John

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/PATCH.java
```diff
@@ -0,0 +1,19 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import javax.ws.rs.HttpMethod;
+
+/**
+ * Indicates that the annotated method responds to HTTP PATCH requests.
```
sparuvu: may be - `Annotation to support the Http PATCH requests.`

docs/source/manual/jaxrs.rst
```diff
@@ -26,6 +26,26 @@ To include the correlation id in the catalina.out file using log4j, update the l
 .. |usageLink| replace:: More on this later
 .. _usageLink: `Usage`_
 
+PATCH
+--------------
+
+`PATCH is a HTTP verb (similar to GET, POST, DELETE, PUT) to push partial changes to the REST resource. JAX-RS by default doesn't support PATCH hence Beadledom adds the support for PATCH explicitly.
```
sparuvu: IMO, it would be good if we could link the PATCH to this wiki article - https://en.wikipedia.org/wiki/Patch_verb

sparuvu:
> @sparuvu comments addressed here. @b-boogaard I went with Sundeeps description of the annotation in the jaxrs.rst. Concise, simple and exactly what the user needs to know. I removed the git url that was linked in the description because don't see how it relates to PATCH at all...

I think I like what Brian has. It is more detailed and more importantly it points that implementing services have the freedom to implement whatever PATCH algorithm they want to go with.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/PATCH.java
EricChristensen: The current description is consistent with other jaxrs http annotation descriptions https://docs.oracle.com/javaee/7/api/javax/ws/rs/PUT.html

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/PATCH.java
EricChristensen: If you insist :) [bb8be52](https://github.com/cerner/beadledom/pull/36/commits/bb8be5203c6721fe2b3ef5d03c017bef69441e43)

EricChristensen:
@sparuvu brian's version of jaxrs.rst [bb8be52](https://github.com/cerner/beadledom/pull/36/commits/bb8be5203c6721fe2b3ef5d03c017bef69441e43)

docs/source/manual/jaxrs.rst
EricChristensen: link to wiki on brian's version of rst [f388ee2](https://github.com/cerner/beadledom/pull/36/commits/f388ee2f8c06434cc427eafd2717d1f38ba67d1d)

checkstyle-suppressions.xml
```diff
@@ -4,4 +4,5 @@
     <suppress files="generated-sources" checks=".*"/>
     <suppress files="generated-test-sources" checks=".*"/>
     <suppress files="com.cerner.beadledom.client.resteasy.ApacheHttpClient4Dot3Engine" checks=".*"/>
+    <suppress files="com.cerner.beadledom.jaxrs.PATCH" checks="AbbreviationAsWordInName" />
```
johnlcox: I just want to confirm this is necessary because checkstyle is expecting the classname to be `Patch`?

If JAX-RS hadn't already set the standard for these annotations we would name it that way, but `PATCH` is correct in following the standard of the other JAX-RS annotations.

EricChristensen: Yep that is the reason.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/PATCH.java
johnlcox: +1 to matching the jaxrs descriptions.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/PATCH.java
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * @author Eric Christensen
```
johnlcox: Add `@since 2.5` below author.

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/PATCHSpec.scala
johnlcox: I think it should be a separate test directly in the `describe("Proxied Clients")` block that tests only that patch works.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/PATCH.java
EricChristensen: [ffcd9d0](https://github.com/cerner/beadledom/pull/36/commits/ffcd9d078905a7bdd6933cff11a5de34ab44e762)

johnlcox:
+1 to @b-boogaard's doc text.

johnlcox:
Will it be desired for beadledom to include a class with additional mediatypes like json-patch and json-merge-patch that can be used similarly to `MediaType.APPLICATION_JSON`?

This questions came to mind while looking at this review, but I think it would be a separate code change if it is desired.

EricChristensen:
@johnlcox we are planning to use json-merge-patch in our application so yes I would say that is a desired behavior.

b-boogaard:
I think adding the media types for patch is a good idea.

EricChristensen:
Is there any reason that adding the media types could not be done in this code change?

johnlcox:
I'd rather see it as a separate code change because I think there needs to be some discussion on the strategy we want to use for supporting additional media types in general. This probably warrants starting with an issue for that discussion prior to any code changes.

johnlcox:
+1

b-boogaard:
+1

johnlcox:
I logged #37 for the media type discussion.

sparuvu:
+1

nimeshsubramanian:
+1

PR #41, opened 2017-08-19
Add exception to request trace error logging

johnlcox:

Added the `JsonProcessingException` to the error log message if serializing the request trace fails to serialize to JSON.

How was it tested?
----

Added a unit test to confirm the exception is logged out.

How to test
----

- [ ] `mvn clean install -U`

Reviewers
----


stagemonitor/src/main/java/com/cerner/beadledom/stagemonitor/request/LogJsonRequestTraceReporter.java
```diff
@@ -35,16 +31,20 @@
...
     this.logger = LoggerFactory.getLogger(LogJsonRequestTraceReporter.class);
+    this.errorLogger =
+        LoggerFactory.getLogger(LogJsonRequestTraceReporter.class.getSimpleName() + ".Error");
     this.objectMapper = objectMapper;
     this.requestMonitorPlugin = Stagemonitor.getPlugin(RequestMonitorPlugin.class);
     this.jsonRequestTraceLoggerPlugin = Stagemonitor.getPlugin(JsonRequestTraceLoggerPlugin.class);
   }
 
   //@VisibleForTesting
   LogJsonRequestTraceReporter(
```
b-boogaard: Am I right in saying, that this second constructor is only used during testing?

We don't need to worry about injecting this second logger in from anywhere do we?

johnlcox: Correct.

This class only gets created here: https://github.com/cerner/beadledom/blob/master/stagemonitor/src/main/java/com/cerner/beadledom/stagemonitor/LogJsonRequestTraceReporterProvider.java

And used here (to register with stagemonitor): https://github.com/cerner/beadledom/blob/master/stagemonitor/src/main/java/com/cerner/beadledom/stagemonitor/RequestMonitorLifecycle.java

stagemonitor/src/main/java/com/cerner/beadledom/stagemonitor/request/LogJsonRequestTraceReporter.java
```diff
@@ -60,10 +60,14 @@ public void reportRequestTrace(ReportArguments reportArguments) throws Exception
...
-        // It's not valid JSON, this log should only contain JSON, so log to error logger instead
+      } catch (JsonProcessingException e) {
+        // Error writing call stack to JSON, this log should only contain JSON,
+        // so log to error logger instead.
         errorLogger
-            .error("Unable to serialize stack request trace to json. Request: {}", requestTrace);
+            .error(
+                "Unable to serialize stack request trace to json. Error: {} Request: {}",
+                e,
+                requestTrace);
```
b-boogaard: Is there anyway get can get an idea of the size of the `requestTrace`? If so, should we either truncate them or not log them past a certain size? If so, should we add a configuration item so consumers can toggle the max size? Or perhaps a configuration option to allow consumers to opt out of logging the `requestTrace` in error scenarios all together?

johnlcox: I think we could do that, but I'd rather take this as a first step and see if we need configuration options.

b-boogaard: Sounds good.

Issue #43, opened 2017-08-22
Catch-All Exception Mapper


cbfshr:
I had a conversation with @johnlcox regarding possibly adding an exception mapper to Beadledom that will handle all uncaught exceptions in a service, map them, and return an error response in JSON format. This is similar to https://github.com/cerner/beadledom/issues/24, except that this issue is more about the exception mapper, which would utilize the format discussed in that issue.

The reason this came up was because all of our services which do not handle exceptions as they're thrown are sending responses in plain text with stack traces, and we don't have any real control over what gets sent back.

John proposed the exception mapper would be a class the implements `ExceptionMapper<Throwable>`.  An example of this is outlined in this article: http://www.codingpedia.org/ama/error-handling-in-rest-api-with-jersey/#22_Uncheckedtechnical_exceptions.

This would hopefully be a first step towards unifying our response structure across application services by implementing the catch all exception mapper with the proper JSON response. I'm opening this issue to discuss design/format/structure of the response, and any feedback is welcomed.

johnlcox:
Fixed by 3ceffecc1f67cb5c4fa7a4cae51174571e6717ac

Issue #24, opened 2017-02-24
Better formatted error messages in service responses


venki09:
Our services currently return responses with 40x-50x if we encounter any client-side/server-side errors. The response now contains a plain text message explaining what's bad in the request and wraps it in to a Jax-rs response object. For example: all 400 responses use the [JaxRsParamCondition](https://github.com/cerner/beadledom/blob/2.4/jaxrs/src/main/java/com/cerner/beadledom/jaxrs/JaxRsParamConditions.java#L44-L50) method provided by beadledom. 

It would be helpful to have a more structured error message [Google API as an example]( https://developers.google.com/drive/v2/web/handle-errors) passed to the service consumer which they can parse and take necessary actions.

b-boogaard:
@venki09 is the idea that beadledom provides a serializable model that supports this format (i.e. a jackson model)?

venki09:
@b-boogaard : Yeah that is the idea. Let me know if there are better ways of handling this. Thanks

b-boogaard:
@venki09 I am pretty sure there are plans for something like this. The blocker is defining and settling on what the structure should actually be.

Nino55N:
We actually ran into this being a problem with the last few releases of hi-service_client: https://jira2.cerner.com/browse/HEALTHEINT-23727. I have proposed that Infra make a small update to hi-service_client in order to get us around this issue for the time being but a services update to return JSON in these instances would be the best solution.

L2lucidream:
+1 to structure google is using, or even a simplified version of it.
```java
{
  "error": {
    "code": 500,
    "message": "my error message" 
  }
}
```

johnlcox:
What I like about Google's error structure is that the `errors` array could be considered optional, so simple error messages could use the structure you have shown above, but something that is performing full validation on inputs may return back a full list of everything that needs to be correct in order to send a valid request.

bjtieman:
Including a correlation id to link the error reported to the client to the error logged by the server would be helpful.  I'm otherwise in favor of the Google API.

johnlcox:
The client should be setting the correlation id header when making the request, so that information should already be available to the client.

johnlcox:
Fixed by 3ceffecc1f67cb5c4fa7a4cae51174571e6717ac

PR #45, opened 2017-08-31
JSON Exception Mappers

cbfshr:
I have modified the existing WebApplicationExceptionMapper and added two new ExceptionMappers (one catch-all and one for Resteasy Internal Exceptions). These ExceptionMappers have responses that follow a Json structure outlined in the [Google API](https://developers.google.com/drive/v2/web/handle-errors). The ExceptionMappers no longer return the exception messages as a part of the response because the message could potentially leak sensitive information such as system information or stack traces. Logging statements have been added to log out the exception information so that it can be properly investigated by the service's owners.

This change addresses https://github.com/cerner/beadledom/issues/24 and https://github.com/cerner/beadledom/issues/43.

How was it tested?
----
This change was tested primarily with unit tests. These unit tests cover the edge cases of the simple ExceptionMapper `toResponse` method. There are also unit tests that simulate a resource throwing an exception, so that an actual response is returned and can be tested. I also consumed these changes in a project that consumes Beadledom and verified that for a variety of thrown exceptions, they are being mapped by these ExceptionMappers and are returned with the expected format and values.

How to test
----
- [ ] `mvn clean install -U`

Reviewers
----


jaxrs/pom.xml
```diff
@@ -53,13 +69,12 @@
             <groupId>javax.ws.rs</groupId>
             <artifactId>javax.ws.rs-api</artifactId>
         </dependency>
-
-        <!--Test Dependencies-->
         <dependency>
             <groupId>org.jboss.resteasy</groupId>
             <artifactId>resteasy-jaxrs</artifactId>
-            <scope>test</scope>
```
shiftyeyes: Since this is a generic jaxrs module, I'm not sure if we want to pull in resteasy libraries. Maybe a jaxrs-resteasy module?

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
```diff
@@ -0,0 +1,70 @@
...
+    logger.error("An unhandled Failure exception was thrown.", exception);
+
+    int statusCode;
+    if (exception.getResponse() != null) {
+      statusCode = exception.getResponse().getStatus();
+    } else {
+      statusCode = INTERNAL_SERVER_ERROR.getStatusCode();
+    }
+
+    if (statusCode >= 400 && statusCode < 500) {
```
shiftyeyes: Since you are already logging the exception on line 42, this can be removed.

jaxrs/pom.xml
```diff
@@ -33,15 +33,31 @@
...
+            <artifactId>auto-value</artifactId>
+        </dependency>
+        <dependency>
             <groupId>com.google.code.findbugs</groupId>
             <artifactId>jsr305</artifactId>
             <scope>provided</scope>
         </dependency>
         <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
```
b-boogaard: We shouldn't need to add a dependency on gson. We already have jackson pulled in, we should be able to just use that.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
```diff
@@ -0,0 +1,70 @@
...
+import org.jboss.resteasy.spi.WriterException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Exception Mapper for all unhandled Exceptions extending from the {@link Failure} hierarchy
+ * (specifically for Resteasy Built-in Internally-Thrown Exceptions like {@link ReaderException},
+ * {@link WriterException}, etc.) thrown by the service. See
+ * http://docs.jboss.org/resteasy/docs/3.1.4.Final/userguide/html_single/index.html#builtinException
+ * for more information.
```
b-boogaard: The way we tend to prefer our javadoc is to have a single summary sentence at the top, then a newline and then any additional documentation.

```java
/**
 * Implementation of {@link ExceptionMapper} for {@link Failure} family of exceptions.
 *
 * <p>{@link Failure} is a family of exceptions that Resteasy will throw when errors occur 
 * internally within the framework. Then intention of this exception mapper is to just restructure
 * the errors into a standard json format.
 *
 * @see <a href="http://docs.jboss.org/resteasy/docs/3.1.4.Final/userguide/html_single/index.html#builtinException">Resteasy Built-in Internally-Thrown Exceptions</a>
```

Or something similar.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * @author Cal Fisher
```
b-boogaard: add `@since 2.6`

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+@Provider
+public class FailureExceptionMapper implements ExceptionMapper<Failure> {
+
+  private static final Logger logger = LoggerFactory.getLogger(FailureExceptionMapper.class);
+
+  /**
+   * Generates a Response object with a Status of the Failure, Content-Type 'application/json', and
```
b-boogaard: Same comment as above, single summary sentence, than continue documentation in a new paragraph.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
```diff
@@ -0,0 +1,70 @@
...
+import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
+
+import com.cerner.beadledom.jaxrs.models.JsonError;
+
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+import javax.ws.rs.ext.Provider;
+
+import org.jboss.resteasy.spi.Failure;
```
johnlcox: Since this class depends on resteasy classes it belongs in the resteasy module. This will resolve @shiftyeyes comment [1] as well.

[1] https://github.com/cerner/beadledom/pull/45/files#r136570400

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
```diff
@@ -0,0 +1,70 @@
...
+  public Response toResponse(Failure exception) {
+
+    logger.error("An unhandled Failure exception was thrown.", exception);
+
+    int statusCode;
+    if (exception.getResponse() != null) {
+      statusCode = exception.getResponse().getStatus();
+    } else {
+      statusCode = INTERNAL_SERVER_ERROR.getStatusCode();
+    }
```
b-boogaard: if we wanted to keep this logic could we condense it a bit?

```java
Response response = exception.getResponse();

int statusCode = response == null ? INTERNAL_SERVER_ERROR.getStatusCode() : response.getStatusCode();
```

or maybe even

```java
int code = e.getResponse() == null ? INTERNAL_SERVER_ERROR.getStatusCode() : e.getResponse().getStatus();
```

I'm not sure what will fit best on a line.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
```diff
@@ -0,0 +1,70 @@
...
+
+    if (statusCode >= 400 && statusCode < 500) {
+      logger.warn("An unhandled WebApplicationException was thrown.", exception);
```
b-boogaard: If we are going to keep this logging logic we shouldn't advertise this as a `WebApplicationException`

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
```diff
@@ -0,0 +1,70 @@
...
+        .status(statusCode)
+        .entity(createJsonError(statusCode))
+        .type(MediaType.APPLICATION_JSON)
+        .build();
+  }
+
+  private static JsonError createJsonError(int statusCode) {
+    return JsonError.builder()
+        .code(statusCode)
+        .message(Response.Status.fromStatusCode(statusCode).getReasonPhrase())
```
b-boogaard: can `fromStatusCode` ever return `null`?

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/ThrowableExceptionMapper.java
```diff
@@ -0,0 +1,51 @@
...
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+import javax.ws.rs.ext.Provider;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Exception Mapper for all unhandled Exceptions thrown by the service.
```
b-boogaard: I don't think `Mapper` or the second `Exception` need to be capitalized.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/ThrowableExceptionMapper.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * @author Brian van de Boogaard
+ * @author Cal Fisher
```
b-boogaard: we should add `@since` tag

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/ThrowableExceptionMapper.java
```diff
@@ -0,0 +1,51 @@
...
+
+    return Response
+        .status(Response.Status.INTERNAL_SERVER_ERROR)
+        .entity(createJsonError())
+        .type(MediaType.APPLICATION_JSON)
+        .build();
+  }
+
+  private static JsonError createJsonError() {
+    return JsonError.builder()
```
b-boogaard: How do people feel about just inlining this above?

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/ThrowableExceptionMapper.java
```diff
@@ -0,0 +1,51 @@
...
+public class ThrowableExceptionMapper implements ExceptionMapper<Throwable> {
+
+  private static final Logger logger = LoggerFactory.getLogger(ThrowableExceptionMapper.class);
+
+  /**
+   * Generates a Response object with a Status of 500, Content-Type 'application/json', and a
+   * JsonError entity containing details about the unhandled exception in Json format.
+   *
+   * @param exception the Throwable exception that was not handled
+   * @return a json response with status 500
```
b-boogaard: should we just say something like

```java
@return a {@link JsonError} containing a {@link Response.Status.INTERNAL_SERVER_ERROR}
```

I'm not actually sure if that second `@link` works or not. 😬

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
johnlcox: Could also go multiline:

```java
int code = e.getResponse() == null 
    ? INTERNAL_SERVER_ERROR.getStatusCode() 
    : e.getResponse().getStatus();
```

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
johnlcox: It seems like this was wanted so client errors (400-499) are logged as warnings while server errors (500-599) are logged as errors.

I've heard before that some 404 exceptions that are getting logged are not desired at all. Does this same sentiment apply to all 400-level errors or is there value in logging out all 400-level errors (at least in this case of a Failure exception)?

A 400 error could occur here if the `MessageBodyReader` fails with an unknown exception or the XML marshaller (which likely wouldn't affect JSON services) throws an unknown exception.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
johnlcox: Yes. if the status code is not a known value in the `Status` enum.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/ThrowableExceptionMapper.java
johnlcox: Do you mean removing this method and putting it in the `.entity()` call above?

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/WebApplicationExceptionMapper.java
```diff
@@ -0,0 +1,69 @@
...
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+import javax.ws.rs.ext.Provider;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A Mapper for {@link WebApplicationException}s that sets the {@code Content-Type} header and
```
johnlcox: These first sentence descriptions for each of the exception mappers should be consistent. It should definitely be called out as an `Exception mapper` or `{@link ExceptionMapper}` rather than just a `Mapper`.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/WebApplicationExceptionMapper.java
```diff
@@ -0,0 +1,69 @@
...
+ * entity body.
+ *
+ * <p>The default handling of {@link WebApplicationException}s, at least in Resteasy, does not
+ * include the exception message or the {@code Content-Type} header. This mapper will set both of
+ * the {@code Content-Type} and entity body if they are not set as well as log the exception
+ * message.
+ *
+ * @author Cal Fisher
```
johnlcox: Add `@since 2.6`

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/exceptionmappers/BaseExceptionMapperSpec.scala
```diff
@@ -0,0 +1,31 @@
...
+  * Base spec to be used by ExceptionMapper spec classes.
+  *
+  * @author Cal Fisher
+  */
+class BaseExceptionMapperSpec
+    extends FunSpec with ShouldMatchers with BeforeAndAfterEach with MockitoSugar {
+
+  val gson = new Gson
+
+  def createJsonError(status: Status): JsValue = {
```
johnlcox: Rather than using this method and using Gson you could have custom error matchers to use in the tests. @b-boogaard already wrote a matcher that we have been using. It's probably not complete for all of the cases needing to be tested with these exception mappers, but it should be easy to expand.

The matcher could go in the `beadledom-testing` module.

```scala
/**
 * [JsonError] matchers.
 *
 * @author Brian van de Boogaard
 */
trait JsonErrorMatchers {
  val mapper = new ObjectMapper()

  class JsonErrorResponse(expectedStatus: Int, expectedMessage: String) extends Matcher[String] {
    def apply(left: String) = {
      val jsonError = mapper.readValue(left, classOf[JsonError])
      val code = jsonError.code()
      val message = jsonError.message()
      MatchResult(
        code == expectedStatus && message.equals(expectedMessage),
        s"""expected: [code: $expectedStatus, message: $expectedMessage] actual [code: $code, message: $message]""",
        s"""expected: [code: $expectedStatus, message: $expectedMessage] matched [code: $code, message: $message]"""
      )
    }
  }

  class BadRequestMatcher(expectedMessage: String) extends JsonErrorResponse(400, expectedMessage)
  class NotFoundMatcher(expectedMessage: String) extends JsonErrorResponse(404, expectedMessage)

  def beBadRequestError(expectedMessage: String = Response.Status.BAD_REQUEST
      .getReasonPhrase) = new BadRequestMatcher(expectedMessage)
  def beNotFoundError(
      expectedMessage: String = Response.Status.NOT_FOUND.getReasonPhrase) = new NotFoundMatcher(
    expectedMessage)
  def beJsonError(expectedStatus: Int, expectedMessage: String) = new JsonErrorResponse(
    expectedStatus, expectedMessage)
}

object JsonErrorMatchers extends JsonErrorMatchers
```

jaxrs/src/test/java/com/cerner/beadledom/jaxrs/provider/FakeDao.java
```diff
@@ -0,0 +1,8 @@
+package com.cerner.beadledom.jaxrs.provider;
+
+public class FakeDao {
```
johnlcox: nit: Can we call this `FakeRepository`?

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/exceptionmappers/BaseExceptionMapperSpec.scala
```diff
@@ -0,0 +1,31 @@
...
+
+import javax.ws.rs.core.Response.Status
+
+/**
+  * Base spec to be used by ExceptionMapper spec classes.
+  *
+  * @author Cal Fisher
+  */
+class BaseExceptionMapperSpec
+    extends FunSpec with ShouldMatchers with BeforeAndAfterEach with MockitoSugar {
```
johnlcox: Use `MustMatchers` instead of `ShouldMatchers.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
johnlcox: Another thought. What might be good to do (but probably as a separate issue/PR) would be to allow configuring a list of status codes to not be logged.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * Generates a Response object with a Status of the Failure, Content-Type 'application/json', and
+   * a JsonError entity containing details about the unhandled exception in Json format.
+   *
+   * @param exception the Failure exception that was not handled
+   * @return a json response with the exception's status or 500 if the exception has no response
+   */
+  @Override
+  public Response toResponse(Failure exception) {
```
shiftyeyes: You aren't using the error code from the Failure. Do we want to propagate that into the json error?

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
```diff
@@ -0,0 +1,70 @@
...
+      statusCode = INTERNAL_SERVER_ERROR.getStatusCode();
+    }
+
+    if (statusCode >= 400 && statusCode < 500) {
+      logger.warn("An unhandled WebApplicationException was thrown.", exception);
+    } else if (statusCode >= 500) {
+      logger.error("An unhandled WebApplicationException was thrown.", exception);
+    }
+
+    return Response
```
shiftyeyes: Since you are creating a new response object, we are loosing headers and other information that may have been on the original response.  You could use `Response.fromResponse();` if the response exists on the Failure object.

core/src/main/java/com/cerner/beadledom/core/BeadledomModule.java
```diff
@@ -44,9 +49,11 @@ protected void configure() {
     install(new StagemonitorModule());
     install(new SwaggerModule());
 
-    bind(PlainTextWebApplicationExceptionMapper.class);
     bind(JsonParseExceptionMapper.class);
     bind(JsonMappingExceptionMapper.class);
+    bind(WebApplicationExceptionMapper.class);
```
shiftyeyes: Since this is a non-passive change, we should do a major version pump.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/ThrowableExceptionMapper.java
b-boogaard: @johnlcox yes.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
b-boogaard: I think we do. I imagine something like

```java
@Provider
public class FailureExceptionMapper implements ExceptionMapper<Failure> {

  @Override
  public Response toResponse(Failure e) {
    Response.Status status = Response.Status.fromStatusCode(e.getErrorCode());

    if (status == null) {
      status = Response.Status.INTERNAL_SERVER_ERROR;
    }

    return Response
        .status(status)
        .type(MediaType.APPLICATION_JSON)
        .entity(
            JsonError.builder()
                .code(status.getStatusCode())
                .message(status.getReasonPhrase()))
        .build();
  }
}
```

but with the logging as well

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapperSpec.scala
```diff
@@ -0,0 +1,197 @@
...
+import org.mockito.Mockito._
+
+import play.api.libs.json.Json
+
+import javax.ws.rs.core.HttpHeaders._
+import javax.ws.rs.core.MediaType
+import javax.ws.rs.core.Response.Status._
+
+/**
+  * Unit tests for {@link FailureExceptionMapper}.
```
b-boogaard: The syntax in scala to link to a class looks like:

```scala
/**
  * Unit tests for [[FailureExceptionMapper]].
  . . .
  */
```

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/exceptionmappers/ThrowableExceptionMapperSpec.scala
```diff
@@ -0,0 +1,138 @@
...
+import org.mockito.Mockito._
+
+import play.api.libs.json.Json
+
+import javax.ws.rs.core.HttpHeaders.CONTENT_TYPE
+import javax.ws.rs.core.MediaType
+import javax.ws.rs.core.Response.Status._
+
+/**
+  * Unit tests for {@link ThrowableExceptionMapper}.
```
b-boogaard: Same comment as [above](https://github.com/cerner/beadledom/pull/45/files/de92be5b169160f9df0a4a700efb175061616c40#r136654766)

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/exceptionmappers/ThrowableExceptionMapperSpec.scala
```diff
@@ -0,0 +1,138 @@
...
+    response = new MockHttpResponse
+  }
+
+  describe("ThrowableExceptionMapper") {
+    describe("unit testing the behavior of ThrowableExceptionMapper class when an exception is " +
+        "thrown from a resource") {
+      val dispatcher = MockDispatcherFactory.createDispatcher()
+      dispatcher.getRegistry.addSingletonResource(fakeResource)
+      dispatcher.getProviderFactory.registerProvider(classOf[JacksonJsonProvider])
+      dispatcher.getProviderFactory.registerProvider(classOf[ThrowableExceptionMapper])
```
b-boogaard: I think you could create the `dispatcher` above with the other `val`s and configure it in a `beforeAll`. Then you could remove the describe that has the really long sentence.

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/exceptionmappers/BaseExceptionMapperSpec.scala
b-boogaard: We tend to prefer `BeforeAndAfter` over `BeforeAndAfterEach` as well. The main disadvantage of `BeforeAndAfter` is that you can't write stackable traits using it (basically a trait that extends `BeforeAndAfter` and calls `super.before`), but we haven't had the need to do that (yet at least). The main win is that it offers a nicer synxax:

```scala
before {
  . . .
}
```
vs.

```scala
override def beforeEach: Unit = {
  . . .
}
```

johnlcox:
Please rebase from master and fix any checkstyle conflicts. I pushed a change that fixes checkstyle to run properly again.

johnlcox:
Update the change log. This would probably be an entry under `Enhancements` indicating it updates the `WebApplicationExceptionMapper` and adds exception mappers for other uncaught error cases.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
cbfshr: What message should be returned in the case where the status code is not a known value? Should the `Response` and the `JsonError` still set the status/code as this unknown value?

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapperSpec.scala
cbfshr: https://github.com/cerner/beadledom/pull/45/commits/865a017a55ce93a2c50c539260b17f6bbcdfe110#diff-0734403a349f9724fd379e4bead2c7e8R20

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/exceptionmappers/ThrowableExceptionMapperSpec.scala
cbfshr: https://github.com/cerner/beadledom/pull/45/commits/865a017a55ce93a2c50c539260b17f6bbcdfe110#diff-538c006eff07d527ef9e8a576ed47cebR19

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
cbfshr: Updated the javadoc: https://github.com/cerner/beadledom/pull/45/commits/865a017a55ce93a2c50c539260b17f6bbcdfe110#diff-94f0fc074c60a4a4d74f16bebf75afc9R17

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
cbfshr: https://github.com/cerner/beadledom/pull/45/commits/865a017a55ce93a2c50c539260b17f6bbcdfe110#diff-94f0fc074c60a4a4d74f16bebf75afc9R26

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
cbfshr: https://github.com/cerner/beadledom/pull/45/commits/865a017a55ce93a2c50c539260b17f6bbcdfe110#diff-94f0fc074c60a4a4d74f16bebf75afc9R34

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/ThrowableExceptionMapper.java
cbfshr: Updated the javadoc: https://github.com/cerner/beadledom/pull/45/commits/865a017a55ce93a2c50c539260b17f6bbcdfe110#diff-8fa995d7fc0ffa3c90069a9fc131c59bR16

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/ThrowableExceptionMapper.java
cbfshr: https://github.com/cerner/beadledom/pull/45/commits/865a017a55ce93a2c50c539260b17f6bbcdfe110#diff-8fa995d7fc0ffa3c90069a9fc131c59bR25

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
cbfshr: https://github.com/cerner/beadledom/pull/45/commits/ec2febdf850acc1686b13715c230407d77733959#diff-94f0fc074c60a4a4d74f16bebf75afc9R44

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
cbfshr: Removed `WebApplicationException` from the log messages: https://github.com/cerner/beadledom/pull/45/commits/ec2febdf850acc1686b13715c230407d77733959#diff-94f0fc074c60a4a4d74f16bebf75afc9R48

cbfshr:
@johnlcox I rebased with master. I have also updated the change log: https://github.com/cerner/beadledom/pull/45/commits/865a017a55ce93a2c50c539260b17f6bbcdfe110#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR3

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/ThrowableExceptionMapper.java
cbfshr: Moved this inline with the `.entity()` call: https://github.com/cerner/beadledom/pull/45/commits/ec2febdf850acc1686b13715c230407d77733959#diff-8fa995d7fc0ffa3c90069a9fc131c59bR46

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/ThrowableExceptionMapper.java
cbfshr: How does this look? https://github.com/cerner/beadledom/pull/45/commits/865a017a55ce93a2c50c539260b17f6bbcdfe110#diff-8fa995d7fc0ffa3c90069a9fc131c59bR36

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/WebApplicationExceptionMapper.java
cbfshr: Updated each of these to say `An {@link ExceptionMapper} for the {@link [enter exception class here]} family of exceptions.`: https://github.com/cerner/beadledom/pull/45/commits/865a017a55ce93a2c50c539260b17f6bbcdfe110#diff-94f0fc074c60a4a4d74f16bebf75afc9R17

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/WebApplicationExceptionMapper.java
cbfshr: https://github.com/cerner/beadledom/pull/45/commits/865a017a55ce93a2c50c539260b17f6bbcdfe110#diff-919ca4a6128e1364a8780c6a64f12dccR23

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
cbfshr: Used the error code from the Failure for the json error following a similar structure to @b-boogaard's suggestion above. Also included logging: https://github.com/cerner/beadledom/pull/45/commits/42cf0d727d134d6218f18ea1118901c89fa3a9d0#diff-94f0fc074c60a4a4d74f16bebf75afc9R44

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/exceptionmappers/ThrowableExceptionMapperSpec.scala
cbfshr: While writing tests, I tend to split them up with `describe` blocks when different types of functionality are happening, so the `describe` block with the really long sentence is meant to indicate that an actual resource is being used. Further down in the spec, you'll see another `describe` block for simply unit testing the class directly. Any suggestions for better naming?

I can move the dispatcher up with the other `val`s like you're suggesting and configure it in a `beforeAll` block, but that would require extending `BeforeAndAfterAll`, which also has the cumbersome syntax: 
```
override def beforeAll(): Unit = {
  ...
}
```

I couldn't find any simpler syntax alternatives (relating back to your suggestion with `BeforeAndAfter` vs `BeforeAndAfterEach`). Would you still prefer I made this change with the `beforeAll`?

jaxrs/src/test/java/com/cerner/beadledom/jaxrs/provider/FakeDao.java
cbfshr: https://github.com/cerner/beadledom/pull/45/commits/b19a43a19a67ff8b1aa1723f780453e7bbeb55ac#diff-eb3b518f028258b8be83878f3ec5ee4eR3

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/exceptionmappers/BaseExceptionMapperSpec.scala
cbfshr: Updated to use `MustMatchers` and `BeforeAndAfter`: https://github.com/cerner/beadledom/pull/45/commits/b19a43a19a67ff8b1aa1723f780453e7bbeb55ac#diff-f4aa1ecb4a6061bdc29737b015c5e9e6R20

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/exceptionmappers/ThrowableExceptionMapperSpec.scala
```diff
@@ -0,0 +1,138 @@
...
+  var response: MockHttpResponse = _
+  val internalServerErrorJsonError = createJsonError(INTERNAL_SERVER_ERROR)
+
+  before {
+    Mockito.reset(fakeRepository)
+    response = new MockHttpResponse
+  }
+
+  describe("ThrowableExceptionMapper") {
+    describe("unit testing the behavior of ThrowableExceptionMapper class when an exception is " +
```
johnlcox: Could this be shorted to something like `when exceptions are thrown from a resource`?

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/exceptionmappers/ThrowableExceptionMapperSpec.scala
johnlcox: Nesting setup in `describe` blocks can result in weird behavior. All code within `describe` blocks, but not within an `it` block will be executed when the spec is being initialized rather than when the nested tests are actually run.

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/exceptionmappers/BaseExceptionMapperSpec.scala
cbfshr: I attempted to add these custom error matchers to the `beadledom-testing` module, but ran into a cyclical dependency issue because the `JsonError` class lives in the `beadledom-jaxrs` module, and the `beadledom-jaxrs` module would be dependent on `beadledom-testing` for these custom error matchers.

Should the `JsonError` and `ErrorDetail` models be moved to a different place?

johnlcox: Good catch.

I had been thinking the `JsonError` and `ErrorDetail` classes should go somewhere else I just hadn't yet formulated where.

I had an offline discussion with Brian and we are thinking they could go in a new module named `beadledom-json-common` with a directory named just `json-common`.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
```diff
@@ -0,0 +1,68 @@
+package com.cerner.beadledom.jaxrs.exceptionmappers;
```
johnlcox: Packages should generally be singular. I think this could either be `jaxrs.exceptionmapper` or simply `jaxrs.exception`.

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/exceptionmappers/ThrowableExceptionMapperSpec.scala
cbfshr: I removed the setup of the dispatcher from the `describe` block to above with the `val`s and inside of a `beforeAll` (for all three specs): https://github.com/cerner/beadledom/pull/45/commits/e1721a2548f63bd69d56eba9dcd838de312ed2d7#diff-0734403a349f9724fd379e4bead2c7e8R32

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/exceptionmappers/ThrowableExceptionMapperSpec.scala
cbfshr: I've shortened this `describe` (for all three specs): https://github.com/cerner/beadledom/pull/45/commits/e1721a2548f63bd69d56eba9dcd838de312ed2d7#diff-0734403a349f9724fd379e4bead2c7e8R47

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
cbfshr: Would you https://github.com/cerner/beadledom/pull/45/commits/3b104b966f9a41e59aa0e0f59a2c3a14a1186db4

jaxrs/pom.xml
cbfshr: Removed depdency on Gson from the parent pom: https://github.com/cerner/beadledom/pull/45/commits/5526d8c027550e6ea5657d742f4ae878f66ac4d7#diff-600376dffeb79835ede4a0b285078036L276 and the jaxrs pom: https://github.com/cerner/beadledom/pull/45/commits/5526d8c027550e6ea5657d742f4ae878f66ac4d7#diff-417137b68e0964b6529eb21151e451c8L45

Instead, using `JsonErrorMatchers` with Jackson: https://github.com/cerner/beadledom/pull/45/commits/5526d8c027550e6ea5657d742f4ae878f66ac4d7#diff-09a50fae7f92d0c897a4330007129927R1

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/exceptionmappers/BaseExceptionMapperSpec.scala
cbfshr: I moved the `JsonError` and `ErrorDetail` classes to a new module named `beadledom-json-common` with a directory name `json-common`: https://github.com/cerner/beadledom/pull/45/commits/b3f0569735e58069bdbc0a3a759fca76a905db1f

I have also added the `JsonErrorMatchers` class above and expanded on the types of exceptions beind tested: https://github.com/cerner/beadledom/pull/45/commits/5526d8c027550e6ea5657d742f4ae878f66ac4d7#diff-09a50fae7f92d0c897a4330007129927R1

Consumed the new `JsonErrorMatchers` in unit tests: https://github.com/cerner/beadledom/pull/45/commits/5526d8c027550e6ea5657d742f4ae878f66ac4d7#diff-dd19a59445cd85b525abe410a27416d4R61

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmapper/FailureExceptionMapper.java
```diff
@@ -0,0 +1,68 @@
+package com.cerner.beadledom.jaxrs.exceptionmapper;
```
johnlcox: Sorry to ask for another change to this package but I actually think `exceptionmapping` would be the best package name.

json-common/src/main/java/com.cerner.beadledom.json.common/models/ErrorDetail.java
```diff
@@ -0,0 +1,88 @@
+package com.cerner.beadledom.json.common.models;
```
johnlcox: Should be singular.

json-common/pom.xml
```diff
@@ -0,0 +1,38 @@
...
+            <artifactId>jsr305</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.wordnik</groupId>
+            <artifactId>swagger-annotations</artifactId>
+        </dependency>
+    </dependencies>
+
+</project>
```
johnlcox: Add a newline.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmapper/WebApplicationExceptionMapper.java
```diff
@@ -0,0 +1,63 @@
+package com.cerner.beadledom.jaxrs.exceptionmapper;
+
+import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
+
+import com.cerner.beadledom.json.common.models.JsonError;
```
johnlcox: Seems like maybe I didn't completely fix checkstyle; the updated Google styleguide only uses two groups (static and normal) with each group ordered alphabetically.

I believe this gist contains an Intellij formatter with the new import order: https://gist.github.com/johnlcox/b7ec8840c74c43df286cbfc52225ed09

cbfshr: I'm assuming this is what you mean:
```
package com.cerner.beadledom.jaxrs.exceptionmapping;

import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;

import com.cerner.beadledom.json.common.model.JsonError;
import javax.ws.rs.WebApplicationException;
import javax.ws.rs.core.MediaType;
import javax.ws.rs.core.Response;
import javax.ws.rs.ext.ExceptionMapper;
import javax.ws.rs.ext.Provider;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
```

Is this the style that every class should follow moving forward, including Scala classes? Or just Java?

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
cbfshr: Modified logic to use `Response.fromResponse(...)` if the response is not null. https://github.com/cerner/beadledom/pull/45/commits/4d7a9e5c99b3678b5e102916ac8844a1eeb86e45

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmapper/FailureExceptionMapper.java
cbfshr: No worries! Changed the package name from `exceptionmapper` to `exceptionmapping`: https://github.com/cerner/beadledom/pull/45/commits/5f20764db12495ac1e321368a52939c7ce24444d

json-common/src/main/java/com.cerner.beadledom.json.common/models/ErrorDetail.java
cbfshr: Changed the package name from `models` to `model`: https://github.com/cerner/beadledom/pull/45/commits/5f20764db12495ac1e321368a52939c7ce24444d

json-common/pom.xml
cbfshr: https://github.com/cerner/beadledom/pull/45/commits/5f20764db12495ac1e321368a52939c7ce24444d#diff-a5a24fe23d11b5098d72bdc07428a727R37

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmapper/FailureExceptionMapper.java
```diff
@@ -55,14 +55,28 @@ public Response toResponse(Failure exception) {
...
-    return Response
-        .status(status)
-        .entity(
-            JsonError.builder()
-                .code(status.getStatusCode())
-                .message(status.getReasonPhrase())
-                .build())
-        .type(MediaType.APPLICATION_JSON)
-        .build();
+    Response response = exception.getResponse();
```
johnlcox: Rather than having mostly the same logic in each of the conditional cases maybe this could be simplified with a private method.

Something like below (note I'm not totally set on this and would like other opinions before this change is made).
```java
return responseWithStatus(exception, status)
     .entity(
         JsonError.builder()
             .code(code)
             .message(status.getReasonPhrase())
             .build())
     .type(MediaType.APPLICATION_JSON)
     .build();


private Response responseWithStatus(Failure e, Response.Status status) {
  return exception.getResponse() != null ? Response.fromResponse(exception.getResponse()) : Response.status(status);
}
```

cbfshr: I agree that this can definitely be condensed a bit. This is also the case for WebApplicationExceptionMapper.

jaxrs/pom.xml
cbfshr: Since this dependency was already here in the test scope for other classes outside the changes for this story, I have moved the FailureExceptionMapper to the `beadledom-resteasy` project: https://github.com/cerner/beadledom/pull/45/commits/ff9039283a54bd60c160857d65055c3ba9d1b35f and moved this dependency back under `Test Dependencies`: https://github.com/cerner/beadledom/pull/45/commits/39df379bb334bb6bda3aa897056d2f9a12a21b52

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
cbfshr: I moved the `FailureExceptionMapper` to the `beadledom-resteasy` module: https://github.com/cerner/beadledom/pull/45/commits/ff9039283a54bd60c160857d65055c3ba9d1b35f

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmapper/WebApplicationExceptionMapper.java
cbfshr: Grouped imports using the provided GoogleStyle formatter: https://github.com/cerner/beadledom/pull/45/commits/6a5509f6056635e81122e800492dfdd530d2dedf

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmappers/FailureExceptionMapper.java
cbfshr: @johnlcox That would be an interesting idea. I'm sure some teams would prefer the option to customize which errors get logged and which do not. If it becomes an issue in the future, we should consider that option, but I think for right now it might not be a bad idea to leave the logging in here. (The previous `PlainTextWebApplicationExceptionMapper` was logging an error every time, so a warning for non-500+ errors would be an improvement nonetheless.)

Regardless: I removed the extra logging code that was left in accidentally: https://github.com/cerner/beadledom/pull/45/commits/865a017a55ce93a2c50c539260b17f6bbcdfe110#diff-94f0fc074c60a4a4d74f16bebf75afc9L42

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmapper/FailureExceptionMapper.java
shiftyeyes: I can't really think of a better method. I do find it strange though that the method is called responseWithStatus and the user passes in a Status, but the Status isn't used if the Failure has a response object.

docs/source/guides/error_handling.rst
```diff
@@ -0,0 +1,118 @@
...
+
+Each of these ExceptionMapper classes has a distinct role in a RESTful service.
+
+The WebApplicationExceptionMapper is intended to handle all exceptions thrown from a service
+that extend from the WebApplicationException hierarchy. This includes exceptions such as
+BadRequestException, NotFoundException, custom exceptions extending from WebApplicationException,
+etc. WebApplicationExceptionMapper treats the exception as-is but sanitizes the exception's
+response, status, and message and formats it as JSON before sending back a response. The status
+will be propagated through to the response as long as it is a valid code; if the code is invalid,
+this will be treated as an Internal Server Error (500). The response message will be derived from
```
venki09: nit: can we say 500 status code instead of 500?

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmapper/FailureExceptionMapper.java
b-boogaard: What about something like:

```java
return Optional.ofNullable(exception.getResponse())
    .map(Response::fromResponse)
    .orElse(Response.status(status))
    .entity(
        JsonError.builder()
            .code(code)
            .message(status.getReasonPhrase())
            .build())
    .type(MediaType.APPLICATION_JSON)
    .build();
```

docs/source/guides/error_handling.rst
```diff
@@ -0,0 +1,118 @@
...
+  - A class that can be used to add additional error details to the JSON response.
+
+These models are serializable objects that can be set as the response's entity body.
+
+Example Response:
+
+.. code-block:: json
+
+  {
+    "errors" [
```
b-boogaard: `"errors": [`

docs/source/guides/error_handling.rst
```diff
@@ -0,0 +1,118 @@
...
+an unhandled exception thrown from a resource endpoint will return an HTTP response in plain text
+or HTML that may contain stack traces or other system-specific information that should not be displayed to the end-user.
+
+Beadledom addresses this concern by implementing exception mappers. By default, Beadledom comes
+packaged with three exception mappers that handle this functionality behind the scenes:
+
+- `WebApplicationExceptionMapper <https://github.com/cerner/beadledom/blob/master/jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmapping/WebApplicationExceptionMapper.java>`_
+  - An exception mapper for exceptions in the WebApplicationException family.
+- `FailureExceptionMapper <https://github.com/cerner/beadledom/blob/master/resteasy/src/main/java/com/cerner/beadledom/resteasy/exceptionmapping/FailureExceptionMapper.java>`_
+  - An exception mapper for exceptions in the Failure family, including Resteasy exceptions.
```
b-boogaard: how about `which encompasses internal exceptions raised by Resteasy`.

docs/source/guides/error_handling.rst
```diff
@@ -0,0 +1,118 @@
...
+- `ThrowableExceptionMapper <https://github.com/cerner/beadledom/blob/master/jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmapping/ThrowableExceptionMapper.java>`_
+  - A catch-all exception mapper for all other unhandled exceptions.
+
+Each of these ExceptionMapper classes has a distinct role in a RESTful service.
+
+The WebApplicationExceptionMapper is intended to handle all exceptions thrown from a service
```
b-boogaard: Should we use the double back-quotes around the class names?

docs/source/guides/error_handling.rst
```diff
@@ -0,0 +1,118 @@
...
+that extend from the WebApplicationException hierarchy. This includes exceptions such as
+BadRequestException, NotFoundException, custom exceptions extending from WebApplicationException,
```
b-boogaard: I'm not sure if we need to mention this or not, but `WebApplicationException` includes some exceptions created by Resteasy, like 405 for example.

docs/source/guides/error_handling.rst
```diff
@@ -0,0 +1,118 @@
...
+Implementation
+______________
+
+Each of the exception mappers mentioned above are readily available for consumption. They
+are automatically added to a project that have installed either the BeadledomModule
+(WebApplicationExceptionMapper and ThrowableExceptionMapper only) or the ResteasyModule
+(WebApplicationExceptionMapper, FailureExceptionMapper, and ThrowableExceptionMapper).
+
+NOTE: If your service is able to throw custom exceptions, it is suggested that a custom
+ExceptionMapper is added to your project as well. This will allow you to customize the response that
```
b-boogaard: Can we make `ExceptionMapper` link here https://docs.jboss.org/resteasy/docs/3.1.4.Final/userguide/html/ExceptionHandling.html#ExceptionMappers ?

b-boogaard: Should we mention to stay consistent with the rest of the error handling `JsonError` should be used for the entity?

b-boogaard:
The rest of my comments are minor and I have confidence in @cbfshr fixes. I am going to go ahead and approve the review so I'm not blocking the merge after everyone else is happy.

json-common/src/main/java/com.cerner.beadledom.json.common/model/ErrorDetail.java
```diff
@@ -0,0 +1,87 @@
+package com.cerner.beadledom.json.common.model;
```
johnlcox: It looks like the package directory structure for the classes in this module got squished into a single directory with dots rather than a directory per package segment.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmapper/FailureExceptionMapper.java
johnlcox: +1 to @b-boogaard suggestion.

docs/source/guides/error_handling.rst
cbfshr: I don't think this is something we need to explicitly state in our documentation. I'll leave it out for the time being.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmapping/WebApplicationExceptionMapper.java
```diff
@@ -0,0 +1,77 @@
...
+   *     {@link JsonError} entity containing details about the unhandled exception in JSON format.
+   */
+  @Override
+  public Response toResponse(WebApplicationException exception) {
+
+    Response response = exception.getResponse();
+    int code = response == null ? INTERNAL_SERVER_ERROR.getStatusCode() : response.getStatus();
+    Response.Status status = Response.Status.fromStatusCode(code);
+
+    if (status == null) {
```
b-boogaard: After some offline conversation I don't think this is the behavior we want. `Response.Status` only supports the core http status codes. It doesn't have representation for any of the extensions. As an example, with the code as it is we would end up mapping `422` to a `500`.

Our real issue here, is when `Response.Status#fromStatusCode` returns `null` we won't have a status phrase to use for the message. We could use `Response.Status#getFamily` and have some message we use depending of the family of the status?

@johnlcox @shiftyeyes

docs/source/guides/error_handling.rst
cbfshr: https://github.com/cerner/beadledom/pull/45/commits/2530c00610b01c4648943d907f7d06846a37a0a4#diff-2275a5c50f335dc3eeff64d39ae09f45R33

docs/source/guides/error_handling.rst
cbfshr: https://github.com/cerner/beadledom/pull/45/commits/2530c00610b01c4648943d907f7d06846a37a0a4#diff-2275a5c50f335dc3eeff64d39ae09f45R68

docs/source/guides/error_handling.rst
cbfshr: Added double back-quotes for all class names: https://github.com/cerner/beadledom/pull/45/commits/2530c00610b01c4648943d907f7d06846a37a0a4#diff-2275a5c50f335dc3eeff64d39ae09f45R74

json-common/src/main/java/com.cerner.beadledom.json.common/model/ErrorDetail.java
cbfshr: Good catch! Made that fix: https://github.com/cerner/beadledom/pull/45/commits/ee2dd85c720353cbdc91bef9862ddd478bfcfcc8

docs/source/guides/error_handling.rst
cbfshr: https://github.com/cerner/beadledom/pull/45/commits/2530c00610b01c4648943d907f7d06846a37a0a4#diff-2275a5c50f335dc3eeff64d39ae09f45R117

docs/source/guides/error_handling.rst
cbfshr: https://github.com/cerner/beadledom/pull/45/commits/2530c00610b01c4648943d907f7d06846a37a0a4#diff-2275a5c50f335dc3eeff64d39ae09f45R80

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmapping/WebApplicationExceptionMapper.java
johnlcox: That sounds good. Use the standard message if one is found; if not fall back to the family and use the family to put a more generic message like `Client Error` and `Server Error`.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmapping/ThrowableExceptionMapper.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+@Provider
+public class ThrowableExceptionMapper implements ExceptionMapper<Throwable> {
+
+  private static final Logger logger = LoggerFactory.getLogger(ThrowableExceptionMapper.class);
+
+  /**
+   * Maps an unhandled {@link Throwable} to a {@link Response}.
+   *
+   * @param exception the {@link Throwable} exception that was not handled.
```
johnlcox: nit: no periods at the end of at clauses. Periods should only be in the summary sentence and description paragraphs.

cbfshr: Removed periods from the end of at clauses: https://github.com/cerner/beadledom/pull/45/commits/cf92faf55674bfa402abe51eff5459464d3ba663#diff-ef962290f5cec93a2a13419308fdf78aR33

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmapper/FailureExceptionMapper.java
cbfshr: Settled on this: https://github.com/cerner/beadledom/pull/45/commits/cf92faf55674bfa402abe51eff5459464d3ba663#diff-a3d460ba2485f948686ff5823fe46af8R49

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmapping/WebApplicationExceptionMapper.java
cbfshr: Added a method called `getMessage(int code)` that will return the message using `Response.Status#fromStatusCode` unless null is returned, then determines the message using `Response.Status#getFamily`: https://github.com/cerner/beadledom/pull/45/commits/cf92faf55674bfa402abe51eff5459464d3ba663#diff-a3d460ba2485f948686ff5823fe46af8R67

PR #46, opened 2017-09-03
Fix checkstyle plugin to run again

johnlcox:

The checkstyle configuration within the pluginManagement section of the
pom was not being used when running the plugin. To get the configuration
to be used, the plugin configuration was removed from pluginManagement
and moved directly to the plugin within the build section.

The checkstyle jar version was also upgraded to 8.1. 8.2 is the latest
version, but it seems to have a bug where it blows up on escaped html
tags like `&lt;` or `&gt;` within javadoc.

How was it tested?
----

Ran `mvn checkstyle:checkstyle` and `mvn clean install`

How to test
----

- [ ] `mvn clean install -U`

Reviewers
----

### DIFF
--- a/google_checks_1.8.xml
+++ b/google_checks_1.8.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <!DOCTYPE module PUBLIC
-        "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
-        "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
+    "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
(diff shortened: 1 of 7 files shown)


google_checks_1.8.xml
```diff
@@ -58,9 +58,7 @@
                       value="LITERAL_TRY, LITERAL_FINALLY, LITERAL_IF, LITERAL_ELSE, LITERAL_SWITCH"/>
         </module>
         <module name="NeedBraces"/>
-        <module name="LeftCurly">
-            <property name="maxLineLength" value="100"/>
```
johnlcox: The line length check is still configured above. This was a property that apparently was deprecated for a long time and was finally removed in checkstyle 8.

PR #57, opened 2017-10-13
Add jooq module with ThreadLocal Guice module

johnlcox:

Adds a new module for using jOOQ for database interactions.

How was it tested?
----

unit tests

How to test
----

- [ ] `mvn clean install -U`

Reviewers
----

### DIFF
--- a/checkstyle-suppressions.xml
+++ b/checkstyle-suppressions.xml
@@ -5,4 +5,9 @@
     <suppress files="generated-test-sources" checks=".*"/>
     <suppress files="com.cerner.beadledom.client.resteasy.ApacheHttpClient4Dot3Engine" checks=".*"/>
     <suppress files="com.cerner.beadledom.jaxrs.PATCH" checks="AbbreviationAsWordInName" />
+
+    <suppress files="com.cerner.beadledom.jooq.DSLContextProvider"
+              checks="AbbreviationAsWordInName"/>
+    <suppress files="com.cerner.beadledom.jooq.DSLContextProviderImpl"
+              checks="AbbreviationAsWordInName"/>
 </suppressions>
--- a/docs/source/manual/index.rst
+++ b/docs/source/manual/index.rst
@@ -19,6 +19,7 @@ User Manual
     Beadledom Jackson <jackson>
     Beadledom JAX-RS <jaxrs>
     Beadledom JAX-RS - Generic Response <jaxrs_generic_response>
+    Beadledom jOOQ <jooq>
     Beadledom Resteasy <resteasy>
     Beadledom Resteasy - Generic Response <resteasy_generic_response>
     Beadledom Stagemonitor <stagemonitor>
--- a/docs/source/manual/jooq.rst
+++ b/docs/source/manual/jooq.rst
@@ -63,7 +63,7 @@ Making a database call:
     }
   }
 
-Configuring the required bindings:
+Below is an example of configuring the required binding uses a C3P0 pooling datasource.
 
 .. code-block:: java
 
--- a/docs/source/manual/jooq.rst
+++ b/docs/source/manual/jooq.rst
@@ -1,0 +1,120 @@
+beadledom-jooq
+==============
+
+Overview
+--------
+
+Beadledom jOOQ is a Google Guice extension for jOOQ for managing the database context and transaction lifecycle. The simplest way to use it is through the use of an annotation and Guice AOP interceptors. This provides functionality similar to `Guice Persist <https://github.com/google/guice/tree/master/extensions/persist>`_ for jOOQ instead of JPA.
+
+See the `jOOQ documentation <https://www.jooq.org/learn/>`_ for details on using and configuring jOOQ.
+
+Download
+--------
+
+Download using Maven
+~~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: xml
+
+  <dependencies>
+      ...
+      <dependency>
+          <groupId>com.cerner.beadledom</groupId>
+          <artifactId>beadledom-jooq</artifactId>
+          <version>[Insert latest version]</version>
+      </dependency>
+      ...
+  </dependencies>
+
+Usage
+-----
+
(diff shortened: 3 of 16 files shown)


docs/source/manual/jooq.rst
```diff
@@ -0,0 +1,120 @@
...
+        dataSource.setMinPoolSize(c3p0Config.getInt("min_size"));
+        dataSource.setMaxPoolSize(c3p0Config.getInt("max_size"));
+        dataSource.setMaxStatementsPerConnection(c3p0Config.getInt("maxStatementsPerConnection"));
+        dataSource.setAcquireIncrement(c3p0Config.getInt("acquire_increment"));
+        dataSource.setTestConnectionOnCheckout(c3p0Config.getBoolean("testConnectionOnCheckout"));
+        dataSource.setTestConnectionOnCheckin(c3p0Config.getBoolean("testConnectionOnCheckin"));
+        dataSource.setMaxIdleTimeExcessConnections(c3p0Config.getInt("maxIdleTimeExcessConnections"));
+
+        return dataSource;
+      } catch (PropertyVetoException e) {
```
shiftyeyes: Could you have the `provideDataSource` throw PropertyVetoException?

checkstyle-suppressions.xml
```diff
@@ -5,4 +5,9 @@
     <suppress files="generated-test-sources" checks=".*"/>
     <suppress files="com.cerner.beadledom.client.resteasy.ApacheHttpClient4Dot3Engine" checks=".*"/>
     <suppress files="com.cerner.beadledom.jaxrs.PATCH" checks="AbbreviationAsWordInName" />
+
+    <suppress files="com.cerner.beadledom.jooq.DSLContextProvider"
```
shiftyeyes: Why not `DslContextProvider`?

docs/source/manual/index.rst
```diff
@@ -19,6 +19,7 @@ User Manual
     Beadledom Jackson <jackson>
     Beadledom JAX-RS <jaxrs>
     Beadledom JAX-RS - Generic Response <jaxrs_generic_response>
+    Beadledom jOOQ <jooq>
```
lal-s: is this a 0?

jooq/src/main/java/com/cerner/beadledom/jooq/JooqTxnInterceptor.java
```diff
@@ -0,0 +1,89 @@
...
+   *
+   * @param originalThrowable the original exception that was thrown and should be propagated
+   * @throws Throwable if an exception occurred while ending the unit of work
+   */
+  private void endUnitOfWork(Throwable originalThrowable) throws Throwable {
+    try {
+      unitOfWork.end();
+    } catch (Throwable t) {
+      if (originalThrowable != null) {
+        throw originalThrowable;
```
shiftyeyes: We are losing all info from `t`. Do we want to log/print it out?

checkstyle-suppressions.xml
johnlcox: Since this class name starts with the name of an existing class from jooq -- `DSLContext` -- we wanted to keep the casing the same even though it's different from how we would traditionally name our classes.

docs/source/manual/index.rst
johnlcox: It's a capital letter o.

docs/source/manual/jooq.rst
johnlcox: I think that should work.

Technically Guice recommends against throwing exceptions in `@Provides` methods at all, but with eager singletons issues will be caught at startup time and these exceptions will be wrapped in a `ProvisionException`.

We had been manually wrapping these in `IllegalStateException` for being consistent and maybe providing some clarity that this really shouldn't have happened.

Does anyone else have an opinion on the current way vs calling out `PropertyVetoException` on the method?

jooq/src/main/java/com/cerner/beadledom/jooq/JooqTxnInterceptor.java
johnlcox: We could log out `t` if the `originalThrowable` is thrown, but will it be confusing to people troubleshooting that two errors occur in the logs where one is likely the underlying cause and the other is a secondary failure during a call to `close()`?

b-boogaard: I think if `originalThrowable` exists here, it is the error you want to focus on. `t` at this point is likely irrelevant since execution is already failing when it occurs. If it is a real issue (which I imagine is the exceptional case) then once the issue that caused `originalThrowable` is resolved `t` should show up as the new `originalThrowable`.

docs/source/manual/jooq.rst
b-boogaard: I don't think guice will let you add it to the method signature. If you want to throw a checked exception from a provider I think you need to use the extension and follow the guidance here: https://github.com/google/guice/wiki/ThrowingProviders

docs/source/manual/jooq.rst
```diff
@@ -0,0 +1,120 @@
...
+
+        ImmutableHierarchicalConfiguration connectionConfig =
+            configuration.immutableConfigurationAt("config.jooq.connection");
+        dataSource.setJdbcUrl(connectionConfig.getString("url"));
+        dataSource.setUser(connectionConfig.getString("username"));
+        dataSource.setPassword(connectionConfig.getString("password"));
+
+        ImmutableHierarchicalConfiguration c3p0Config =
+            configuration.immutableConfigurationAt("config.jooq.c3p0");
+        dataSource.setMinPoolSize(c3p0Config.getInt("min_size"));
```
shiftyeyes: The configuration that is being read is not documented anywhere. Also, I would have a default value if one does not exist in the config.

docs/source/manual/jooq.rst
johnlcox: I will keep it as is then.

docs/source/manual/jooq.rst
johnlcox: This is just documentation of an example of configuring a datasource. Other datasources and/or pools can be used besides C3P0, and can be configured however services prefer.

I updated to clarify that this is just an example: https://github.com/cerner/beadledom/commit/4861647139709047ffe28615902ce61ba8a28013#diff-2a3927f61725d9baa48fb9c016e79882L66

PR #61, opened 2017-11-10
Upgrade modules to Java 1.8

johnlcox:

Make Java 1.8 the minimum Java version so we no longer have a mismatch.

How was it tested?
----

Existing unit tests and integration tests.

How to test
----

- [ ] `mvn clean install -U`

Reviewers
----

### DIFF
--- a/client/pom.xml
+++ b/client/pom.xml
@@ -26,12 +26,6 @@
         <module>beadledom-client-test</module>
     </modules>
 
-    <properties>
-        <java.version>1.6</java.version>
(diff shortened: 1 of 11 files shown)


client/pom.xml
```diff
@@ -26,12 +26,6 @@
         <module>beadledom-client-test</module>
```
shiftyeyes: Update changelog

johnlcox: https://github.com/cerner/beadledom/commit/6074c6f1778bae267458eb98fcd61f6f2ffb61dd

PR #80, opened 2018-01-18
Add dependency-check plugin

johnlcox:

Adds the `dependency-check-plugin` to scan dependencies for OWASP issues. Also updates the dependencies that have fixes and suppresses the ones that are false positives.

How was it tested?
----

Ran `mvn clean install site`

How to test
----

- [ ] `mvn clean install -U`

Reviewers
----

### DIFF
--- a/dependency-check-suppressions.xml
+++ b/dependency-check-suppressions.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.1.xsd">
+  <!-- This CVE is regarding the javascript not the Java implementation -->
   <suppress>
     <notes><![CDATA[
    file name: compiler-0.9.5.jar
@@ -7,6 +8,7 @@
     <gav regex="true">^com\.github\.spullara\.mustache\.java:compiler:.*$</gav>
     <cve>CVE-2015-8862</cve>
   </suppress>
+  <!-- This CVE is regarding xzgrep not the Java implementation -->
   <suppress>
     <notes><![CDATA[
    file name: xz-1.8.jar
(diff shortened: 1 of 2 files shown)


dependency-check-suppressions.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.1.xsd">
+  <suppress>
```
johnlcox: Should we put a comment explaining the reason for each suppression?

b-boogaard: I think that’s a good idea.

pom.xml
```diff
@@ -451,6 +451,11 @@
         <artifactId>stagemonitor-web</artifactId>
         <version>${stagemonitor.version}</version>
       </dependency>
+      <dependency>
+        <groupId>org.tukaani</groupId>
+        <artifactId>xz</artifactId>
```
b-boogaard: Interesting name. I think we should start naming our artifacts like this. ‘Just add a dependency on ab3 and you’re good to go’

pom.xml
```diff
@@ -645,6 +668,22 @@
           </dependency>
         </dependencies>
       </plugin>
+      <plugin>
+        <groupId>org.owasp</groupId>
+        <artifactId>dependency-check-maven</artifactId>
+        <version>3.1.0</version>
+        <configuration>
+          <failBuildOnAnyVulnerability>true</failBuildOnAnyVulnerability>
```
b-boogaard: +1

pom.xml
johnlcox: I think you're on to something here. Apparently `xz` is a compression algorithm so maybe the letters actually mean something.

dependency-check-suppressions.xml
johnlcox: https://github.com/cerner/beadledom/pull/80/commits/cf6b74240c7390911c6ba7c4d8489d593d9e78ac

PR #81, opened 2018-02-08
Update archetype and documentation for new recommended client structure.

johnlcox:

Update archetype and documentation for new recommended client structure.

The new recommend client structure is to have the API interface defined
in the client for automatic proxying via beadledom-client. This differs
from the previous guidance which was to have this interface in the api
project to be shared between the client and service. Sharing the
interface causes issues due to the fact that the service side
implementation can have only one method per path/method combo, making
it very hard to passively add support for new query parameters. By
separating the client and service implementations, the service can
add query parameters as needed and the client can add new methods
to the proxied interface for different variations of query parameters.

How was it tested?
----

Generated a new project from the archetype and verified it matches the new recommendations.
Generated and viewed the updated documentation site.

How to test
----

- [ ] `mvn clean install -U`

Reviewers
----


johnlcox:
Closing this for now. There are some other changes I want to make and then I will reopen a PR.

PR #95, opened 2018-07-10
Support building with JDK 10+

johnlcox:

Support building with JDK 10+, while still building to target Java 1.8.

How was it tested?
----

Built with both JDK8 and JDK10. Included in red-tape and built/ran all tests.

How to test
----

- [ ] `mvn clean install -U`


jaxrs-genericresponse/pom.xml
```diff
@@ -16,6 +16,10 @@
       <groupId>javax.ws.rs</groupId>
       <artifactId>javax.ws.rs-api</artifactId>
     </dependency>
+    <dependency>
+      <groupId>javax.xml.bind</groupId>
+      <artifactId>jaxb-api</artifactId>
```
nathanschile: This is needed because it was broken out with java 9/10?

johnlcox: Yep. JDK 9 has removed jaxb-api from the core library.

PR #96, opened 2018-07-11
Change the HttpClient  to only retry on 503 response codes

### LINKED ISSUE #44: Client Retries on 500 Errors is weird
The default HttpClient retry behavior is to only retry on 503s. We override that to be any 500s which is kind of weird and possibly dangerous when talking about PUT, POST, DELETE methods. We may want to reconsider this and only retry on 503's or at least only for GETS if we do 500s.

johnlcox:

Fixes #44.

Change client retries to only retry 503 status codes.

How was it tested?
----

Ran existing unit and integration tests.

How to test
----

- [ ] `mvn clean install -U`

### DIFF
--- a/client/resteasy-client/src/test/scala/com/cerner/beadledom/client/resteasy/http/DefaultServiceUnavailableRetryStrategySpec.scala
+++ b/client/resteasy-client/src/test/scala/com/cerner/beadledom/client/resteasy/http/DefaultServiceUnavailableRetryStrategySpec.scala
@@ -1,6 +1,4 @@
-package com.cerner.beadledom.client.common.http
-
-import com.cerner.beadledom.client.resteasy.http.DefaultServiceUnavailableRetryStrategy
+package com.cerner.beadledom.client.resteasy.http
 
 import org.apache.http.client.ServiceUnavailableRetryStrategy
 import org.apache.http.client.protocol.HttpClientContext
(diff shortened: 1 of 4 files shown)


client/resteasy-client/src/test/scala/com/cerner/beadledom/client/resteasy/http/DefaultServiceUnavailableRetryStrategySpec.scala
```diff
@@ -26,19 +24,19 @@ class DefaultServiceUnavailableRetryStrategySpec extends FunSpec with Matchers w
...
 
       describe("with status code 500 and executionCount 2") {
-        it should behave like retryableRequest(retryStrategy, 500, 2)
+        it should behave like nonRetryableRequest(retryStrategy, 500, 2)
       }
 
       describe("with status code 599 and executionCount 0") {
-        it should behave like retryableRequest(retryStrategy, 599, 0)
+        it should behave like nonRetryableRequest(retryStrategy, 599, 0)
       }
```
nathanschile: Probably need tests for `describe("with status code 503 and executionCount 2")` and `describe("with status code 503 and executionCount 3")`

johnlcox: https://github.com/cerner/beadledom/pull/96/commits/b682f04ded7a42813b08f18966bb3730bb2ca729

PR #98, opened 2018-07-12
Fix HealthChecker to log dependency exceptions

johnlcox:

Fix `HealthChecker` to actually log dependency exceptions. Previously it was passing the `Optional` to the logger rather than the actual exception.

How was it tested?
----

Existing unit/integration tests.

How to test
----

- [ ] `mvn clean install -U`

### DIFF
--- a/health/service/src/main/java/com/cerner/beadledom/health/internal/HealthChecker.java
+++ b/health/service/src/main/java/com/cerner/beadledom/health/internal/HealthChecker.java
@@ -31,6 +31,7 @@ import org.slf4j.LoggerFactory;
  */
 public class HealthChecker {
   private static final Logger LOGGER = LoggerFactory.getLogger(HealthChecker.class);
+  private static final Range<Integer> HEALTHY_STATUS_RANGE = Range.closed(200, 299);
 
   private final UriInfo uriInfo;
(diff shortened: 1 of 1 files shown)


health/service/src/main/java/com/cerner/beadledom/health/internal/HealthChecker.java
```diff
@@ -31,6 +31,7 @@
  */
 public class HealthChecker {
```
nathanschile: Does a test need to be updated to test that your changes work?

johnlcox: How can we test loggers?

johnlcox: Existing tests cover the other refactorings I did, but I don't know how to test a log message.

johnlcox: The [slf4j-test](http://projects.lidalia.org.uk/slf4j-test) project looks interesting for this, but we will lose our regular log output during tests if we use it.

nathanschile: don't worry about it.

PR #100, opened 2018-07-16
Fix health DTOs Jackson deserialization

nathanschile:

* Fixes health DTOs Jackson deserialization by adding @JsonDeserialize and @JsonCreator

How was it tested?
----

* Unit test

How to test
----

- [ ] `mvn clean install -U`

### DIFF
--- a/health/api/src/main/java/com/cerner/beadledom/health/dto/BuildDto.java
+++ b/health/api/src/main/java/com/cerner/beadledom/health/dto/BuildDto.java
@@ -1,7 +1,9 @@
 package com.cerner.beadledom.health.dto;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.fasterxml.jackson.annotation.JsonView;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
 import com.google.auto.value.AutoValue;
 import com.wordnik.swagger.annotations.ApiModel;
 import com.wordnik.swagger.annotations.ApiModelProperty;
@@ -15,8 +17,9 @@ import java.util.Optional;
  *
  * @since 1.4
  */
-@ApiModel(description = "Indicates the list of dependencies of this service.")
 @AutoValue
+@ApiModel(description = "Indicates the list of dependencies of this service.")
+@JsonDeserialize(builder = AutoValue_BuildDto.Builder.class)
 public abstract class BuildDto {
   /**
    * Creates a new builder for {@code BuildDto}.
@@ -41,22 +44,31 @@ public abstract class BuildDto {
   public abstract Optional<String> getBuildDateTime();
 
   @AutoValue.Builder
+  @JsonPOJOBuilder(withPrefix = "set")
   public abstract static class Builder {
+
+    @JsonCreator
+    private static BuildDto.Builder create() {
+      return BuildDto.builder();
+    }
 
     abstract Builder setArtifactName(Optional<String> artifactName);
 
+    @JsonProperty("artifactName")
     public Builder setArtifactName(String artifactName) {
       return setArtifactName(Optional.ofNullable(artifactName));
     }
 
     abstract Builder setVersion(Optional<String> version);
 
+    @JsonProperty("version")
     public Builder setVersion(String version) {
       return setVersion(Optional.ofNullable(version));
     }
 
     abstract Builder setBuildDateTime(Optional<String> buildDateTime);
 
+    @JsonProperty("buildDateTime")
     public Builder setBuildDateTime(String buildDateTime) {
       return setBuildDateTime(Optional.ofNullable(buildDateTime));
     }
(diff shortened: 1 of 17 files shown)


health/api/src/main/java/com/cerner/beadledom/health/dto/BuildDto.java
```diff
@@ -15,8 +17,9 @@
  *
  * @since 1.4
  */
-@ApiModel(description = "Indicates the list of dependencies of this service.")
 @AutoValue
+@ApiModel(description = "Indicates the list of dependencies of this service.")
+@JsonDeserialize(builder = AutoValue_BuildDto.Builder.class)
```
johnlcox: `BuildDto.Builder.class`

health/api/src/main/java/com/cerner/beadledom/health/dto/BuildDto.java
```diff
@@ -15,8 +17,9 @@
  *
  * @since 1.4
  */
-@ApiModel(description = "Indicates the list of dependencies of this service.")
 @AutoValue
+@ApiModel(description = "Indicates the list of dependencies of this service.")
```
johnlcox: Make sure the order of these annotations is consistent for all the health DTOs. It seems like most had `JsonDeserialize` before `ApiModel`.

health/api/src/test/scala/com/cerner/beadledom/health/dto/HealthDependenciesDtoSpec.scala
```diff
@@ -0,0 +1,22 @@
...
+                         .setName("name")
+                         .setMessage("message")
+                         .setPrimary(true)
+                         .setLinks(LinksDto.builder().setSelf("link").build())
+                         .build()
+
+  var healthDependenciesDto = HealthDependenciesDto.builder()
+                              .setDependencies(List(healthDependency).asJava).build()
+
+  it must behave like serializableByJackson(healthDependenciesDto)
```
johnlcox: Seeing this now in an actual `behave` statement, I wonder if the behavior method would be better named as `aJacksonSerializableObject`

health/api/src/main/java/com/cerner/beadledom/health/dto/BuildDto.java
nathanschile: Updated, https://github.com/cerner/beadledom/pull/100/commits/5b462705934358b648b7e9f6b532f7364b1d0aa6

health/api/src/main/java/com/cerner/beadledom/health/dto/BuildDto.java
nathanschile: Fixed ordering, https://github.com/cerner/beadledom/pull/100/commits/16bb80cf2621d60970f0ad16e06a7708768dd997

health/api/src/test/scala/com/cerner/beadledom/health/dto/HealthDependenciesDtoSpec.scala
nathanschile: I like it, https://github.com/cerner/beadledom/pull/100/commits/ac509b10974ebd5fe955358ff52a3000348ee107

health/api/src/main/java/com/cerner/beadledom/health/dto/BuildDto.java
```diff
@@ -41,22 +44,31 @@ public static Builder builder() {
...
   public abstract static class Builder {
 
+    @JsonCreator
+    private static BuildDto.Builder create() {
+      return BuildDto.builder();
+    }
+
     abstract Builder setArtifactName(Optional<String> artifactName);
 
+    @JsonProperty("artifactName")
```
sparuvu: why do we need to give the property name explicitly. I thought Jackson automatically infers from the method name.

health/api/src/test/scala/com/cerner/beadledom/health/dto/HealthDependencyDtoSpec.scala
```diff
@@ -53,4 +54,16 @@ class HealthDependencyDtoSpec extends FunSpec with MustMatchers {
...
+    .setHealthy(true)
+    .setId("1")
+    .setName("name")
+    .setMessage("message")
+    .setPrimary(true)
+    .setLinks(LinksDto.builder().setSelf("link").build())
+    .build()
+
+  it must behave like aJacksonSerializableObject(healthDependencyDto)
+
```
sparuvu: nit: empty line here and at other places.

health/api/src/main/java/com/cerner/beadledom/health/dto/BuildDto.java
johnlcox: The Beadledom default object mapper uses underscores, but the health check needs to use camel case.

health/api/src/test/scala/com/cerner/beadledom/health/dto/HealthDependencyDtoSpec.scala
nathanschile: Removed empty lines, https://github.com/cerner/beadledom/pull/100/commits/e90e469d9c9a4c65fb23ae4bbd01e37a90a39527

PR #110, opened 2018-08-27
Handle jax-rs responses that do not contain a entity (#109)

### LINKED ISSUE #109: [beadledom-jaxrs-clientproxy] Handle responses that do not contain a body
### Describe the problem:

When using GenericResponse with the beandledom-jaxrs-clientproxy project, and the resource interface returns a GenericResponse without a type, the client call fails with an exception because the code expects the response to have a entity.

#### Steps to reproduce:

  1. Have a interface method that returns a untyped GenericResponse. `GenericResponse getGenericWithNoType();`
  2. Attempt to call this method using a GenericClientProxy 

#### Observed Results:

  * Exception is thrown
```

java.lang.Class cannot be cast to java.lang.reflect.ParameterizedType
java.lang.ClassCastException: java.lang.Class cannot be cast to java.lang.reflect.ParameterizedType
	at com.cerner.beadledom.client.proxy.GenericClientProxy.buildGenericResponse(GenericClientProxy.java:56)
	at com.cerner.beadledom.client.proxy.GenericClientProxy.invoke(GenericClientProxy.java:42)
```

#### Expected Results:

  * A GenericResponse is returned without a `body()`

nathanschile:

Fixes #109 by handling jax-rs respones that do not have a entity.

### How was it tested?

Unit tested

### How to test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Beadledom Changelog
 
+## 3.1 - TBD
+
+### Breaking Changes
+
+### Enhancements
+
+### Defects Corrected
+* Fixed beadledom client deserialization to GenericResponse when the JAX-RS response has no entity.
+
 ## 3.0 - 3 August 2018
 
 ### Breaking Changes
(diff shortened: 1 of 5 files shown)


nathanschile:
A workaround is to give a `Void` return type. e.g. `GenericResponse<Void> getGeneric();`

johnlcox:
Update changelog?

PR #112, opened 2018-09-05
Add offset based pagination

### LINKED ISSUE #51: Offset based pagination component
In a conversation with @johnlcox we have discussed a possibility of adding a common offset based pagination component to the beadledom. The component would decorate a list of objects returned from target methods with all necessary pagination elements.
 
Using John’s implementation of pagination as part of one of the Cerner projects as a model the component would include:

- a subclass of WriterInterceptor applied to target methods with a specific return type (e.g. 
   OffsetPaginatedList<T>)
- a response DTO decorated by the interceptor with:
    Items – list of items of type T
    totalResults – total number of results across all the pages
    hasMore - an indication wheter there are additional items on another page
    firstLink – link to the first page
    lastLink – link to the last page
    nextLink – link to a page after current (null if current is the last page)
    prevLink – link to a page before current (null if current is the first page)
 
I'm opening this issue to discuss design of the solution and feedback would be appreciated.

ikottman:

Add module which provides offset pagination query parameters and hypermedia links. Fixes #51 

### How was it tested?

Unit and integration tests, as well as integrated into an existing service and manually verified pagination links are added to responses as expected. 

### How to test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,13 @@
 
 ## 3.1 - TBD
 
-### Additions
-* Add Offset based pagination with beadledom-pagination
+### Breaking Changes
+
+### Enhancements
+* Add offset based pagination with beadledom-pagination.
+
+### Defects Corrected
+* Fixed beadledom client deserialization to GenericResponse when the JAX-RS response has no entity.
 
 ## 3.0 - 3 August 2018
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,13 @@
 
 ## 3.1 - TBD
 
+### Additions
+* Add offset based pagination with beadledom-pagination.
+
 ### Breaking Changes
 
 ### Enhancements
-* Add offset based pagination with beadledom-pagination.
+
 
 ### Defects Corrected
 * Fixed beadledom client deserialization to GenericResponse when the JAX-RS response has no entity.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Beadledom Changelog
 
+## 3.1 - TBD
+
+### Additions
+* Add Offset based pagination with beadledom-pagination
+
 ## 3.0 - 3 August 2018
 
 ### Breaking Changes
@@ -18,11 +23,6 @@ swagger 2 and OpenAPI 3.
 
 ### Defects Corrected
 * Fixed health DTOs Jackson deserialization.
-
-## 2.9 - September 5 2018
-
-### Additions
-* Add Offset based pagination with beadledom-pagination 
 
 ## 2.8 - 16 May 2018
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,11 @@ swagger 2 and OpenAPI 3.
 ### Defects Corrected
 * Fixed health DTOs Jackson deserialization.
 
+## 2.9 - September 5 2018
+
+### Additions
+* Add Offset based pagination with beadledom-pagination 
+
 ## 2.8 - 16 May 2018
 
 ### Additions
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -18,6 +18,7 @@ This file contains the list of developers who contributed to this repository
 * Bryan Baugher [@bbaugher][bryan-baugher]
 * Eric Christensen [@EricChristensen][eric-christensen]
 * Cal Fisher [@cbfshr][cal-fisher]
+* Ian Kottman [@ikottman][ian-kottman]
 
 [john-leacox]: https://github.com/johnlcox
 [jacob-williams]: https://github.com/brokensandals
@@ -33,5 +34,6 @@ This file contains the list of developers who contributed to this repository
 [bryan-baugher]: https://github.com/bbaugher
 [eric-christensen]: https://github.com/EricChristensen
 [cal-fisher]: https://github.com/cbfshr
+[ian-kottman]: https://github.com/ikottman
 
 Thanks to everyone who helped in building this repository.
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -19,6 +19,7 @@ This file contains the list of developers who contributed to this repository
 * Eric Christensen [@EricChristensen][eric-christensen]
 * Cal Fisher [@cbfshr][cal-fisher]
 * Ian Kottman [@ikottman][ian-kottman]
+* Will Pruyn [@lulzWill][will-pruyn]
 
 [john-leacox]: https://github.com/johnlcox
 [jacob-williams]: https://github.com/brokensandals
@@ -35,5 +36,6 @@ This file contains the list of developers who contributed to this repository
 [eric-christensen]: https://github.com/EricChristensen
 [cal-fisher]: https://github.com/cbfshr
 [ian-kottman]: https://github.com/ikottman
+[will-pruyn]: https://github.com/lulzWill
 
 Thanks to everyone who helped in building this repository.
--- a/client/beadledom-client-example/example-api/src/main/java/com/cerner/beadledom/client/example/PaginatedClientResource.java
+++ b/client/beadledom-client-example/example-api/src/main/java/com/cerner/beadledom/client/example/PaginatedClientResource.java
@@ -4,7 +4,6 @@ import com.cerner.beadledom.client.example.model.JsonOne;
 import com.cerner.beadledom.jaxrs.GenericResponse;
 import com.cerner.beadledom.pagination.models.OffsetPaginatedListDto;
 
-import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
@@ -20,6 +19,6 @@ public interface PaginatedClientResource {
   @GET
   @Produces("application/json")
   GenericResponse<OffsetPaginatedListDto<JsonOne>> index(
-      @DefaultValue("0") @QueryParam("offset") final Long offset,
-      @DefaultValue("20") @QueryParam("limit") final Integer limit);
+      @QueryParam("offset") final Long offset,
+      @QueryParam("limit") final Integer limit);
 }
--- a/client/beadledom-client-example/example-api/src/main/java/com/cerner/beadledom/client/example/PaginatedClientResource.java
+++ b/client/beadledom-client-example/example-api/src/main/java/com/cerner/beadledom/client/example/PaginatedClientResource.java
@@ -1,8 +1,7 @@
 package com.cerner.beadledom.client.example;
 
-import com.cerner.beadledom.client.example.model.JsonOne;
+import com.cerner.beadledom.client.example.model.JsonOneOffsetPaginatedListDto;
 import com.cerner.beadledom.jaxrs.GenericResponse;
-import com.cerner.beadledom.pagination.models.OffsetPaginatedListDto;
 
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
@@ -18,7 +17,7 @@ import javax.ws.rs.QueryParam;
 public interface PaginatedClientResource {
   @GET
   @Produces("application/json")
-  GenericResponse<OffsetPaginatedListDto<JsonOne>> index(
+  GenericResponse<JsonOneOffsetPaginatedListDto> index(
       @QueryParam("offset") final Long offset,
       @QueryParam("limit") final Integer limit);
 }
--- a/client/beadledom-client-example/example-api/src/main/java/com/cerner/beadledom/client/example/PaginatedClientResource.java
+++ b/client/beadledom-client-example/example-api/src/main/java/com/cerner/beadledom/client/example/PaginatedClientResource.java
@@ -1,0 +1,25 @@
+package com.cerner.beadledom.client.example;
+
+import com.cerner.beadledom.client.example.model.JsonOne;
+import com.cerner.beadledom.jaxrs.GenericResponse;
+import com.cerner.beadledom.pagination.models.OffsetPaginatedListDto;
+
+import javax.ws.rs.DefaultValue;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
+
+/**
+ * Proxy client interface for PaginatedResource
+ *
+ * @author Ian Kottman
+ */
+@Path("/paginated")
+public interface PaginatedClientResource {
+  @GET
+  @Produces("application/json")
+  GenericResponse<OffsetPaginatedListDto<JsonOne>> index(
+      @DefaultValue("0") @QueryParam("offset") final Long offset,
+      @DefaultValue("20") @QueryParam("limit") final Integer limit);
+}
--- a/client/beadledom-client-example/example-api/src/main/java/com/cerner/beadledom/client/example/PaginatedClientResource.java
+++ b/client/beadledom-client-example/example-api/src/main/java/com/cerner/beadledom/client/example/PaginatedClientResource.java
@@ -11,7 +11,7 @@ import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
 
 /**
- * Proxy client interface for PaginatedResource
+ * Proxy client interface for PaginatedResource.
  *
  * @author Ian Kottman
  */
--- a/client/beadledom-client-example/example-api/src/main/java/com/cerner/beadledom/client/example/PaginatedClientResource.java
+++ b/client/beadledom-client-example/example-api/src/main/java/com/cerner/beadledom/client/example/PaginatedClientResource.java
@@ -2,7 +2,6 @@ package com.cerner.beadledom.client.example;
 
 import com.cerner.beadledom.client.example.model.JsonOneOffsetPaginatedListDto;
 import com.cerner.beadledom.jaxrs.GenericResponse;
-
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
--- a/client/beadledom-client-example/example-api/src/main/java/com/cerner/beadledom/client/example/model/OffsetPaginatedListDto.java
+++ b/client/beadledom-client-example/example-api/src/main/java/com/cerner/beadledom/client/example/model/OffsetPaginatedListDto.java
@@ -1,0 +1,60 @@
+package com.cerner.beadledom.client.example.model;
+
+import com.cerner.beadledom.pagination.OffsetPaginatedList;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * Offset based pagination list.
+ *
+ * <p>Represents an offset based paginated list in a rest response. Is used on the client side
+ * instead of {@link OffsetPaginatedList} to circumvent type erasure.
+ *
+ */
+public interface OffsetPaginatedListDto<T> {
+
+  @JsonProperty("totalResults")
+  @Nullable
+  Long totalResults();
+
+  @JsonProperty("firstLink")
+  String firstLink();
+
+  @JsonProperty("lastLink")
+  @Nullable
+  String lastLink();
+
+  @JsonProperty("prevLink")
+  @Nullable
+  String prevLink();
+
+  @JsonProperty("nextLink")
+  @Nullable
+  String nextLink();
+
+  @JsonProperty("items")
+  List<T> items();
+
+  interface Builder<B extends Builder<B, T>, T> {
+
+    @JsonProperty("items")
+    B items(List<T> items);
+
+    @JsonProperty("totalResults")
+    B totalResults(Long totalResults);
+
+    @JsonProperty("firstLink")
+    B firstLink(String firstLink);
+
+    @JsonProperty("lastLink")
+    B lastLink(String lastLink);
+
+    @JsonProperty("nextLink")
+    B nextLink(String nextLink);
+
+    @JsonProperty("prevLink")
+    B prevLink(String prevLink);
+  }
+}
--- a/client/beadledom-client-example/example-api/src/main/java/com/cerner/beadledom/client/example/model/OffsetPaginatedListDto.java
+++ b/client/beadledom-client-example/example-api/src/main/java/com/cerner/beadledom/client/example/model/OffsetPaginatedListDto.java
@@ -11,7 +11,6 @@ import javax.annotation.Nullable;
  *
  * <p>Represents an offset based paginated list in a rest response. Is used on the client side
  * instead of {@link OffsetPaginatedList} to circumvent type erasure.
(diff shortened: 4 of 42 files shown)


client/beadledom-client-example/example-api/src/main/java/com/cerner/beadledom/client/example/PaginatedClientResource.java
```diff
@@ -0,0 +1,25 @@
...
+ * Proxy client interface for PaginatedResource.
+ *
+ * @author Ian Kottman
+ */
+@Path("/paginated")
+public interface PaginatedClientResource {
+  @GET
+  @Produces("application/json")
+  GenericResponse<OffsetPaginatedListDto<JsonOne>> index(
+      @DefaultValue("0") @QueryParam("offset") final Long offset,
```
nathanschile: nit. Remove `@DefaultValue` since it has no affect on the JAX-RS client.

client/beadledom-client-test/src/test/scala/com/cerner/beadledom/client/ClientServiceSpec.scala
```diff
@@ -66,6 +69,38 @@ class ClientServiceSpec(contextRoot: String, servicePort: Int)
...
 
+    describe("Pagination") {
+      it("injects pagination links to the response") {
+        val injector = getInjector(List(new ResourceOneModule))
+
+        val paginatedResource = injector.getInstance(classOf[PaginatedClientResource])
+
+        val results : OffsetPaginatedListDto[JsonOne] = paginatedResource.index(0L, 10).body()
+
+        results mustNot be(null)
```
nathanschile: If you add 
`results.items().get(0).getHello mustBe "Hello World"`
the test will fail with `java.lang.ClassCastException: java.util.LinkedHashMap cannot be cast to com.cerner.beadledom.client.example.model.JsonOne` that is because Jackson doesn't know how to deserialize the json to java objects [1]. We either need to figure out a way to get TypeReference to be used, or remove the examples from the client since it doesn't work.

[1] https://www.baeldung.com/jackson-collection-array

CHANGELOG.md
```diff
@@ -19,6 +19,11 @@ swagger 2 and OpenAPI 3.
 ### Defects Corrected
 * Fixed health DTOs Jackson deserialization.
 
+## 2.9 - September 5 2018
```
nathanschile: Should this go with the 3.1 release? We could backport it to the 2.x line if needed.

johnlcox: I had suggested using `@since 2.9` in the code since we would likely backport it, but I now realize the changelog complicates things a little bit.

CONTRIBUTORS.md
```diff
@@ -18,6 +18,7 @@ This file contains the list of developers who contributed to this repository
 * Bryan Baugher [@bbaugher][bryan-baugher]
 * Eric Christensen [@EricChristensen][eric-christensen]
 * Cal Fisher [@cbfshr][cal-fisher]
+* Ian Kottman [@ikottman][ian-kottman]
```
johnlcox: Should @lulzwill be added too?

pagination/src/main/java/com/cerner/beadledom/pagination/parameters/PaginationParameters.java
```diff
@@ -0,0 +1,45 @@
...
+ * Query parameters for an endpoint with offset based pagination.
+ *
+ * <p>This class can be added with the @BeanParam annotation to add the query parameters needed
+ * to support offset based pagination. Allows for injection of default values for limit and offset.
+ * *NOTE* This will not work with configured field names.
+ *
+ * @author Will Pruyn
+ * @since 2.9
+ */
+public class PaginationParameters {
```
johnlcox: `OffsetPaginationParameters`

pagination/src/main/java/com/cerner/beadledom/pagination/parameters/PaginationParameters.java
```diff
@@ -0,0 +1,45 @@
...
+
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.UriInfo;
+
+/**
+ * Query parameters for an endpoint with offset based pagination.
+ *
+ * <p>This class can be added with the @BeanParam annotation to add the query parameters needed
+ * to support offset based pagination. Allows for injection of default values for limit and offset.
+ * *NOTE* This will not work with configured field names.
```
johnlcox: Might be good to have the note in it's own paragraph so it stands out a bit more.

pagination/src/main/java/com/cerner/beadledom/pagination/parameters/PaginationParameters.java
```diff
@@ -0,0 +1,45 @@
...
+ * @since 2.9
+ */
+public class PaginationParameters {
+  @Context
+  UriInfo uriInfo;
+
+  /**
+   * Retrieves the value of the limit field from the request. Will use the configured field name
+   * to find the parameter.
+   * @return The value of the limit.
```
johnlcox: Add a line break before the at-params.

pagination/src/main/java/com/cerner/beadledom/pagination/parameters/PaginationParameters.java
```diff
@@ -0,0 +1,45 @@
...
+   */
+  public Integer getLimit() {
```
johnlcox: Add `@Nullable` annotation for both of these methods.

pagination/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+
+    <artifactId>beadledom-pagination</artifactId>
+    <name>Beadledom Pagination</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.google.auto.value</groupId>
+            <artifactId>auto-value</artifactId>
+            <scope>provided</scope>
+        </dependency>
```
nathanschile: Add dependencies

```
    <dependency>
      <groupId>com.google.auto.value</groupId>
      <artifactId>auto-value-annotations</artifactId>
    </dependency>

<dependency>
  <groupId>javax.inject</groupId>
  <artifactId>javax.inject</artifactId>
</dependency>


```

pagination/src/main/java/com/cerner/beadledom/pagination/BeadledomPaginationModule.java
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * <p>Module provides the configurations needed to use offset based pagination. This includes
+ * the {@link OffsetPaginatedListLinksWriterInterceptor} as well as default values for the limit and
+ * offset fields and their respective field names. The default values for the limit, offset, and
+ * their respective field names can be configured using one of the provided constructors.
+ *
+ * @author Will Pruyn
+ * @since 2.9
+ */
+public class BeadledomPaginationModule extends AbstractModule {
```
nathanschile: Rename to BeadledomOffsetPaginationModule

CONTRIBUTORS.md
ikottman: Yep, I'll add him.

pagination/src/main/java/com/cerner/beadledom/pagination/parameters/PaginationParameters.java
ikottman: Nice catch, I'll rename it.

CHANGELOG.md
ikottman: Yeah I was unsure what to do with the changelog. Is what I have here acceptable or do we need to do something else?

pagination/src/main/java/com/cerner/beadledom/pagination/BeadledomPaginationModule.java
ikottman: Will future pagination live in a separate module? I was under the impression if we added cursor based pagination it would be in this module.

pagination/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+            <artifactId>beadledom-json-common</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>javax.ws.rs</groupId>
+            <artifactId>javax.ws.rs-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>javax.validation</groupId>
+            <artifactId>validation-api</artifactId>
+            <version>2.0.1.Final</version>
```
johnlcox: Can this version be moved up to the parent pom to be managed in one place?

pagination/src/main/java/com/cerner/beadledom/pagination/BeadledomPaginationModule.java
johnlcox: Could be shortened to just `OffsetPaginationModule`.

pagination/src/main/java/com/cerner/beadledom/pagination/OffsetPaginatedList.java
```diff
@@ -0,0 +1,94 @@
...
+     *
+     * @param limitFieldName the limit field name used to create the page.
+     * @param limit the limit used to create the page.
+     * @param offsetFieldName the offset field name used to create the page.
+     * @param offset the offset used to create the page.
+     * @param totalResults the total results available.
+     * @param hasMore indicates if there are more results available.
+     * @return The {@link Builder} being used for the {@link OffsetPaginatedList}.
+     */
+    public OffsetPaginatedList.Builder<T> metadata(
```
johnlcox: Should all of these parameters be marked as nullable?

pagination/src/main/java/com/cerner/beadledom/pagination/OffsetPaginatedList.java
```diff
@@ -0,0 +1,94 @@
+package com.cerner.beadledom.pagination;
+
+import com.cerner.beadledom.pagination.parameters.LimitParameter;
+import com.cerner.beadledom.pagination.parameters.OffsetParameter;
+
```
johnlcox: Looks like your formatter is not following the latest google style guide which no longer groups imports.

pagination/src/main/java/com/cerner/beadledom/pagination/parameters/LimitParameter.java
```diff
@@ -0,0 +1,98 @@
...
+ *
+ * @author Brian van de Boogaard
+ * @author Will Pruyn
+ * @since 2.9
+ */
+@ApiModel
+public class LimitParameter extends AbstractParameter<Integer> {
+
+  @ApiModelProperty(value = "Total number of items to return in the response.", dataType = "int",
+      allowableValues = "range[0, 100]")
```
johnlcox: The swagger doc will be wrong if non-default values are used.

pagination/src/main/java/com/cerner/beadledom/pagination/parameters/LimitParameter.java
```diff
@@ -0,0 +1,98 @@
...
+    Integer limit;
+    try {
+      limit = Integer.parseInt(this.limit);
+    } catch (NumberFormatException e) {
+      throw InvalidParameterException.create(
+          "Invalid type for '" + this.getParameterFieldName() + "': " + this.limit
+              + " - int is required.");
+    }
+
+    if (limit < 0 || limit > 100) {
```
johnlcox: This won't work with a non-default value.

CHANGELOG.md
nathanschile: @johnlcox and I talked. Lets have this enhancement go out on 3.1 release of Beadledom. Going from 2.x to 3.x is trivial, http://engineering.cerner.com/beadledom/3.0/docs/releases/Beadledom30.html

pagination/src/main/java/com/cerner/beadledom/pagination/OffsetPaginatedList.java
nathanschile: Here is the updated formatter, https://github.cerner.com/hi-infra/appservices-styleguide/tree/updated-google-style

client/beadledom-client-example/example-service/src/main/java/com/cerner/beadledom/client/example/PaginatedResourceImpl.java
```diff
@@ -0,0 +1,34 @@
...
+  public GenericResponse<OffsetPaginatedList<JsonOne>> index(
+      @BeanParam OffsetPaginationParameters paginationParams) {
+    List<JsonOne> items = Lists.newArrayList();
+    for (int i = 0; i < 1000; i++) {
+      items.add(JsonOne.create("One", "Hello World"));
+    }
+
+    OffsetPaginatedList<JsonOne> body = OffsetPaginatedList.<JsonOne>builder()
+        .items(items)
+        .metadata(paginationParams.getLimit(), paginationParams.getOffset(), 1000L, null)
```
nathanschile: This could be updated to be `.metadata(1000L)`. We really don't need to pass the limit and offset everywhere since the OffsetPaginatedListLinksWriterInterceptor has access to the UriInfo.

pagination/src/main/java/com/cerner/beadledom/pagination/OffsetPaginatedList.java
ikottman: Updated with the latest https://plugins.jetbrains.com/plugin/8527-google-java-format , let me know if that was what you were looking for

nathanschile: Didn't know about the plugin, awesome! It still doesn't do import order correctly though, https://github.com/google/google-java-format#intellij 

```
The import ordering is not handled by this plugin, unfortunately. To fix the import order, download the IntelliJ Java Google Style file and import it into File→Settings→Editor→Code Style.
```

I installed the plugin and the Google code style, and everything seems good now. Thanks for the info @ikottman

ikottman: Whoops, totally missed that. I updated my checkstyle rules and re-optimized imports, _now_ the imports should be correct.

pagination/src/main/java/com/cerner/beadledom/pagination/parameters/LimitParameter.java
ikottman: This and your comment on the limit check raises an interesting question. Our swagger doc and the validations only mention the acceptable range of values provided by the caller, which isn't currently configurable. The configurable piece is the default value when nothing is provided by the caller. That means the default value could be configured to be outside the range of acceptable values, and everything will function fine.

For example, you can configure the default limit to be `1,000`, and it will succeed and the prev/next links will be correct etc. But if the caller follows the next link and actually _specifies_ `1,000` for the limit it will throw an error due to it being above `100`. 

All this to say: should the minimum limit, maximum limit, and minimum offset be configurable? If yes, the documentation cannot actually mention those limits since the annotations can't include variables. It would also add even more constructors to the pagination module.

client/beadledom-client-example/example-service/src/main/java/com/cerner/beadledom/client/example/PaginatedResourceImpl.java
ikottman: I added an another convenience method to the builder for this scenario.

pagination/src/main/java/com/cerner/beadledom/pagination/parameters/LimitParameter.java
ikottman: After some offline discussion it was decided to allow the maximum limit to be configurable.

ikottman:
@johnlcox @nathanschile I have responded to all comments, this is ready for review again.

client/beadledom-client-example/example-api/src/main/java/com/cerner/beadledom/client/example/PaginatedClientResource.java
```diff
@@ -0,0 +1,23 @@
+package com.cerner.beadledom.client.example;
+
+import com.cerner.beadledom.client.example.model.JsonOneOffsetPaginatedListDto;
+import com.cerner.beadledom.jaxrs.GenericResponse;
+
```
nathanschile: nit. remove blank line

client/beadledom-client-example/example-api/src/main/java/com/cerner/beadledom/client/example/model/OffsetPaginatedListDto.java
```diff
@@ -0,0 +1,60 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * Offset based pagination list.
+ *
+ * <p>Represents an offset based paginated list in a rest response. Is used on the client side
+ * instead of {@link OffsetPaginatedList} to circumvent type erasure.
+ *
```
venki09: nit: this blank line can be removed

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Breaking Changes
 
 ### Enhancements
+* Add offset based pagination with beadledom-pagination.
```
johnlcox: Since this is a whole new maven module this should probably go under an Additions header.

pagination/src/main/java/com/cerner/beadledom/pagination/OffsetPaginationModule.java
```diff
@@ -0,0 +1,141 @@
...
+ * the {@link OffsetPaginatedListLinksWriterInterceptor} as well as default values for the limit and
+ * offset fields and their respective field names. The default values for the limit, offset, and
+ * their respective field names can be configured using one of the provided constructors.
+ *
+ * @author Will Pruyn
+ * @author Ian Kottman
+ * @since 3.1
+ */
+public class OffsetPaginationModule extends AbstractModule {
+  private final Integer maxLimit;
```
venki09: Is it ok to have state for a guice module class? Would it be better to have this come in through a configuration object?

pagination/src/main/java/com/cerner/beadledom/pagination/parameters/AbstractParameter.java
```diff
@@ -0,0 +1,106 @@
...
+    return Response.status(
+        Response.Status.BAD_REQUEST)
+        .type(MediaType.APPLICATION_JSON)
+        .entity(getError(e))
+        .build();
+  }
+
+  protected JsonError getError(InvalidParameterException e) {
+    return JsonError.builder()
+        .code(400)
```
venki09: nit: Can this be `Response.Status.BAD_REQUEST` as well

pagination/src/main/java/com/cerner/beadledom/pagination/OffsetPaginationModule.java
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  /**
+   * Constructor that enables setting of default values for the offset and limit fields.
+   * Default field names will be set using Beadledom defaults.
+   * @param defaultLimit the default value for the limit field; 20 if null.
+   * @param maxLimit the default value for the limit field; 100 if null.
+   * @param defaultOffset the default value for the offset field; 0 if null.
+   */
+  public OffsetPaginationModule(Integer defaultLimit, Integer maxLimit, Long defaultOffset) {
```
kyleroush: Dont know if you guys thought about using autovalue for this class also so that we could have a builder instead of a bunch of constructors? 
or if you want to keep the constructors have you thought about instead of setting the params in each one just passing the params along to one so you dont have to have all the default log checking in each one.

pagination/src/main/java/com/cerner/beadledom/pagination/models/OffsetPaginatedListDto.java
```diff
@@ -0,0 +1,97 @@
...
+@JsonPropertyOrder({
+    "items",
+    "totalResults",
+    "firstLink",
+    "lastLink",
+    "nextLink",
+    "prevLink"
+})
+@JsonDeserialize(builder = OffsetPaginatedListDto.Builder.class)
+public abstract class OffsetPaginatedListDto<T> {
```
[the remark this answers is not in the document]
ikottman: Looking over the dependencies, there aren't a huge number of service-specific ones that would concern me, especially since currently the client would have to do a concrete implementation of the service anyway, so they end up having similar dependencies.

pagination/src/main/java/com/cerner/beadledom/pagination/OffsetPaginationModule.java
ikottman: While I think it's ok to have state in the module, this has gotten a bit out of hand. I like the idea of a config object, I'll work on adding one.

pagination/src/main/java/com/cerner/beadledom/pagination/parameters/exceptions/InvalidParameterException.java
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Exception used to communicate an invalid parameter.
+ *
+ * <p>Contains metadata used to build {@link Error} and {@link ErrorDetail} for the resulting
+ * JSON response.
+ *
+ * @author Brian van de Boogaard
+ * @since 3.1
+ */
+public final class InvalidParameterException extends RuntimeException {
```
johnlcox: Since we may want to eventually move the `AbstractParameter` into another module for more reusability, what do you think about making `AbstractParameter` package-private and moving this exception up to the same package as `AbstractParameter` and making this exception class package-private as well?

That could give us more freedom to move things around in the future.

pagination/src/main/java/com/cerner/beadledom/pagination/OffsetPaginationModule.java
```diff
@@ -0,0 +1,83 @@
...
+  public OffsetPaginationModule() {
+    this.config = OffsetPaginationConfiguration.builder().build();
+  }
+
+  /**
+   * Override Beadledom defaults.
+   */
+  public OffsetPaginationModule(OffsetPaginationConfiguration config) {
+    this.config = config;
+  }
```
venki09: I feel like a cleaner approach would be to have a requireBinding for this OffsetPaginationConfiguration which the consumer would have to provide. This way, the module doesn't have to worry about handing defaults vs overrides.

I think you just need to add the following in the configure() method below
```
@Override
protected void configure() {
  requireBinding(OffsetPaginationConfiguration.class);
  ........
  ........
}
```

This way we can get rid off the constructors altogether.  If you choose to do this, it would also require a README update.

pagination/src/main/java/com/cerner/beadledom/pagination/parameters/exceptions/InvalidParameterException.java
ikottman: Sounds, good. Will do that now.

pagination/src/main/java/com/cerner/beadledom/pagination/OffsetPaginationModule.java
johnlcox: Might be able to use an optional binding for this with a default.

venki09: Even better!

PR #113, opened 2018-09-12
Prepare of 3.1 release

venki09:
* 3.1 release prepare
* Update release docs.

### How to test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,12 @@
 # Beadledom Changelog
 
-## 3.1 - TBD
+## 3.1 - 12 September 2018
 
 ### Additions
 * Add offset based pagination with beadledom-pagination.
 
-### Breaking Changes
-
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,14 +1,12 @@
...
+## 3.1 - 12 September 2018
 
 ### Additions
 * Add offset based pagination with beadledom-pagination.
 
-### Breaking Changes
-
 ### Enhancements
-
+* Supports offset based pagination.
```
johnlcox: This is already under additions.

venki09: Isn't this also an enhancement? Or maybe a better question. When does something get added to an enhancement?

johnlcox: I think the way we have been doing it is that enhancements are new features on an existing thing and additions are totally new things.

Issue #117, opened 2018-09-18
Handle 0 value to LimitParameter gracefully by throwing BadRequest


venki09:
* Device: Macbook pro 2017
  * OS version: 10.12.6 (High sierra)
  * Java SDK version: 1.8
  * Beadledom version: 3.1

### Step 3: Describe the problem:
* When a service call is made with limit parameter as 0, it is expected to return a BAD_REQUEST response. Instead INTERNAL_SERVER_ERROR is being returned.

#### Observed Results:

```
java.lang.ArithmeticException: / by zero
	at com.cerner.beadledom.pagination.OffsetPaginationLinks.lastLink(OffsetPaginationLinks.java:86) ~[beadledom-pagination-3.1.jar:na]
at com.cerner.beadledom.pagination.OffsetPaginatedListLinksWriterInterceptor.aroundWriteTo(OffsetPaginatedListLinksWriterInterceptor.java:39) ~[beadledom-pagination-3.1.jar:na]
at org.jboss.resteasy.core.interception.AbstractWriterInterceptorContext.proceed(AbstractWriterInterceptorContext.java:129) ~[resteasy-jaxrs-3.6.1.Final.jar:3.6.1.Final]
at org.jboss.resteasy.plugins.interceptors.encoding.GZIPEncodingInterceptor.aroundWriteTo(GZIPEncodingInterceptor.java:92) ~[resteasy-jaxrs-3.6.1.Final.jar:3.6.1.Final]
at org.jboss.resteasy.core.interception.AbstractWriterInterceptorContext.proceed(AbstractWriterInterceptorContext.java:129) ~[resteasy-jaxrs-3.6.1.Final.jar:3.6.1.Final]
at org.jboss.resteasy.plugins.interceptors.encoding.ServerContentEncodingAnnotationFilter.aroundWriteTo(ServerContentEncodingAnnotationFilter.java:60) ~[resteasy-jaxrs-3.6.1.Final.jar:3.6.1.Final]
```

#### Expected Results:

* Return a BAD_REQUEST response.

johnlcox:
Fixed by #118

PR #118, opened 2018-09-18
Handle 0 for Limit parameter gracefully

venki09:

Fixing https://github.com/cerner/beadledom/issues/117 

### How was it tested?

* Unit tests
* Ran the example service against it.

### How to test


client/beadledom-client-test/src/test/scala/com/cerner/beadledom/client/ClientServiceSpec.scala
```diff
@@ -132,6 +132,17 @@ class ClientServiceSpec(contextRoot: String, servicePort: Int)
...
+
+      it("rejects 0 for limit") {
+        val injector = getInjector(List(new ResourceOneModule))
+
+        val paginatedResource = injector.getInstance(classOf[PaginatedClientResource])
+
+        val results: GenericResponse[JsonOneOffsetPaginatedListDto] = paginatedResource.index(1L, 0)
+
+        results mustNot be(null)
+        results.getStatus mustBe 400
```
nathanschile: I've seen instances where people set the limit to 0 when they just want to get a total results count. Why not still have a 200 when limit is 0?

venki09: https://github.com/cerner/beadledom/pull/118/commits/0fcf057b54008675463e241ffa2050c9d8855181 I have made the changed for it. With this, the result would be a single page with empty results block and totalResults would be the total number of results. And everything except firstLink will be null.  Something like this
```
{
  "items": [],
  "totalResults": 3,
  "firstLink": "http://<service-resource-URI>?offset=0&limit=0"
}
```
Also note that the offset will always be 0 since this is the first page (and the only page). Let me know if this is the acceptable behavior

johnlcox:
The more time I've had to think about this the more I think the default should be to disallow 0 for limit and then allow a configuration flag for allowing 0 limit.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 ### Additions
 * Add OffsetPaginatedListDto interface to the pagination module.
+* Add configuration to allow 0 as a pagination limit value (disabled by default - limit=1 as minimum value).
+
+### Defects Corrected
```
johnlcox: Not sure why this is showing as an addition here. If I do a local `git diff` against master it shows that this was already there and is not an addition from this branch. I think github is diffing this incorrectly.

pagination/src/main/java/com/cerner/beadledom/pagination/models/OffsetPaginationConfiguration.java
```diff
@@ -37,6 +37,11 @@
    */
   public abstract String offsetFieldName();
 
+  /**
+   * Flag to denote whether limit value of 0 is allowed. Defaults to false.
+   */
+  public abstract boolean allowZeroLimit();
```
johnlcox: @venki09 I switched this from `Boolean` to `boolean` since it can't be nullable, but I wanted to double check that there wasn't a reason I'm missing for it being a big boolean?

CHANGELOG.md
venki09: I think github UI doesn't work well if we do merge without rebase. As was done in this commit https://github.com/cerner/beadledom/pull/118/commits/cffffa1b60e18286d92dcbf1bcf3c5c72d8976c9

pagination/src/main/java/com/cerner/beadledom/pagination/models/OffsetPaginationConfiguration.java
venki09: No there was no reason for it. I think it makes sense to have it boolean

pagination/src/main/java/com/cerner/beadledom/pagination/parameters/LimitParameter.java
```diff
@@ -68,10 +59,18 @@ protected Integer parse(String param) throws InvalidParameterException {
               + " - int is required.");
     }
 
-    if (limit < 0 || limit > maxLimit) {
```
kyleroush: Would something like this be more readable?
```java
int minlimit = (offsetPaginationConfiguration.allowZeroLimit()? 0:1;
if ( limit < minLimit)
        || limit > offsetPaginationConfiguration.maxLimit()) {
      throw InvalidParameterException.create(
          "Invalid value for '" + this.getParameterFieldName() + "': " + this.limit
              + "  - value between "+minlimit+" and " + offsetPaginationConfiguration.maxLimit()
              + " is required.");
    }
```

johnlcox: I think it does. I'm going to do that and extract it into a private method as well.

johnlcox: https://github.com/cerner/beadledom/pull/118/commits/a3acf4d6055c5c2ad2a61dcd75c4fe21077d12bf

PR #123, opened 2018-11-05
Changed primary health check to only check dependencies that are mark…

nab0310:

The primary health check was changed from checking all dependencies to just checking primary ones. This is the way the health checks were originally intended to work. I also made some changes to the testing framework for the service project. 

### How was it tested?

`mvn clean install`

### How to test

@nathanschile 
@johnlcox


health/service/src/main/java/com/cerner/beadledom/health/HealthModule.java
```diff
@@ -116,8 +116,6 @@ protected void configure() {
     Multibinder<Module> jacksonModuleBinder = Multibinder.newSetBinder(binder(), Module.class);
     jacksonModuleBinder.addBinding().to(Jdk8Module.class);
     jacksonModuleBinder.addBinding().to(JavaTimeModule.class);
-
-    install(MultibindingsScanner.asModule());
```
nab0310: This was incorporated into default Guice as of 4.4. https://google.github.io/guice/api-docs/4.2/javadoc/com/google/inject/multibindings/MultibindingsScanner.html

health/service/src/main/java/com/cerner/beadledom/health/HealthDependency.java
```diff
@@ -9,7 +9,7 @@
  */
 public abstract class HealthDependency {
 
-  private Boolean primary = false;
+  private Boolean primary = true;
```
nab0310: This is a change from defaulting the HealthDependencies from secondary to primary. Because we don't know if our consumers rely on the previous behavior (of `meta/health` checking dependencies if they were uncategorized) one of the solutions we came up with was to make the dependencies default to primary if unspecified. This change preserves the functionality of `meta/health` for projects that don't specify dependency  importance but I'm not sure that it is the correct way to handle this. Feedback would be appreciated.

health/service/src/main/java/com/cerner/beadledom/health/internal/HealthChecker.java
```diff
@@ -56,8 +57,13 @@
    * Other metadata fields on the HealthDto will also be populated.
    */
   public HealthDto doPrimaryHealthCheck() {
-    // TODO - allow some dependencies to be marked as 'secondary', and skip those here
-    return checkHealth(healthDependencies.values());
+    List<HealthDependency> dependencies = new ArrayList<>();
```
nathanschile: You can use the stream api,

```
List<HealthDependency> primaryHealthDependencies = healthDependencies.values().stream()
        .filter(HealthDependency::getPrimary)
        .collect(Collectors.toList());
```

Another option would be to create to create two maps in the constructor, one with the primary and the other with the ancillary dependencies.

health/service/src/test/java/com/cerner/beadledom/health/HealthServiceModule.java
```diff
@@ -0,0 +1,36 @@
...
+  public void configure() {
+    install(new HealthModule());
+  }
+
+  @Provides
+  ServiceMetadata provideServiceMetadata() {
+    return ServiceMetadata.builder()
+        .setBuildInfo(
+            BuildInfo.builder()
+                .setArtifactId("Lombre")
```
nathanschile: Lets use more realistic values

health/service/src/test/scala/com/cerner/beadledom/health/resource/BaseSpec.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  val mustacheFactory = new DefaultMustacheFactory("com/cerner/beadledom/health")
+
+  protected var primaryHealthyDependency: HealthDependency = createDependency(Primitives.generateString(), healthy = true, primary = true)
+  protected var ancillaryHealthyDependency: HealthDependency = createDependency(Primitives.generateString(), healthy = true, primary = false)
+  protected var primaryUnhealthyDependency: HealthDependency = createDependency(Primitives.generateString(), healthy = false, primary = false)
+  protected var ancillaryUnhealthyDependency: HealthDependency = createDependency(Primitives.generateString(), healthy = false, primary = false)
+
+
+  def createDependency(name: String, healthy: java.lang.Boolean, primary: Boolean): HealthDependency =
```
nathanschile: `healthy` can also be a Scala Boolean

testing/src/main/scala/com/cerner/beadledom/testing/Primitives.scala
```diff
@@ -0,0 +1,20 @@
...
+
+import org.scalacheck.Gen
+import org.scalacheck.rng.Seed
+
+/**
+  * Data generation for primitive types.
+  *
+  * @author Nick Behrens
+  */
+
```
nathanschile: nit. Extra space

health/service/src/test/scala/com/cerner/beadledom/health/resource/DependenciesResourceImplSpec.scala
```diff
@@ -1,59 +1,52 @@
...
-  val mustacheFactory = new DefaultMustacheFactory("com/cerner/beadledom/health")
-  val mockUriInfo = mock[UriInfo]
-  val mockUriBuilder = mock[UriBuilder]
-  val url = URI.create("demoUrl")
-  val healthModule = new AbstractModule {
+  * Spec tests for DependenciesResourceImplSpec.
+  *
+  * @author Nimesh Subramanian
+  */
+
```
nathanschile: nit. Remove empty line.

health/service/pom.xml
```diff
@@ -16,6 +16,10 @@
       <groupId>com.cerner.beadledom</groupId>
```
nathanschile: Add changes to CHANGELOG.md, add yourself to CONTRIBUTORS.md

health/service/src/main/java/com/cerner/beadledom/health/internal/HealthChecker.java
nab0310: Changed to use stream api: https://github.com/nab0310/beadledom/commit/4928608b1f95af3516d4c682c8c88b17c0832d15

health/service/src/test/scala/com/cerner/beadledom/health/resource/BaseSpec.scala
nab0310: Changed to a scala boolean: https://github.com/nab0310/beadledom/commit/4928608b1f95af3516d4c682c8c88b17c0832d15

testing/src/main/scala/com/cerner/beadledom/testing/Primitives.scala
nab0310: removed this file in a future commit

health/service/pom.xml
nab0310: Added myself here: https://github.com/cerner/beadledom/pull/123/commits/bd033376b739f9d4baae3808ec2f105145d1f7c3

Do you think these changes would be classified as a defect?

health/service/src/test/scala/com/cerner/beadledom/health/resource/DependenciesResourceImplSpec.scala
nab0310: removed here:https://github.com/cerner/beadledom/pull/123/commits/bd033376b739f9d4baae3808ec2f105145d1f7c3

CHANGELOG.md
```diff
@@ -9,9 +9,7 @@
 
 ### Defects Corrected
 * Fix isHealthy field being included in the health dependency list JSON.
-
-### Defects Corrected
-* Fix isHealthy field being included in the health dependency list JSON.
+* Primary Health Check checking all dependencies.
```
nathanschile: I would give more detail of the changes. Something like

```
* Health resource (/health) now excludes non-primary health dependencies. The non-primary dependencies are still included in the diagnostic resource (/health/diagnostic). The default for health dependencies is also changed to be primary.
```

health/service/pom.xml
nathanschile: I would keep it as a defect since we shouldn't have been returning non-primary dependencies from the health resource.

CHANGELOG.md
nab0310: Changed it to your suggestion here: https://github.com/cerner/beadledom/pull/123/commits/7e25d081e7a14dc0701cb9573ab0645fbddd6414

health/service/src/test/java/com/cerner/beadledom/health/HealthServiceModule.java
nab0310: Changed the values here: 82d60f08015ffdfdef327b1e929f9dc901c880b3

health/service/src/main/java/com/cerner/beadledom/health/HealthDependency.java
johnlcox: This is basically passive.

I think the only change in behavior will be for any dependency implementations that explicitly set `primary` to false, they will actually be excluded from the health check. Arguably this was their intent anyway.

johnlcox: Would it be beneficial to add an ADR directory and include a small ADR documenting our decision on this?

health/service/src/main/java/com/cerner/beadledom/health/internal/HealthChecker.java
```diff
@@ -56,8 +57,10 @@
    * Other metadata fields on the HealthDto will also be populated.
    */
   public HealthDto doPrimaryHealthCheck() {
-    // TODO - allow some dependencies to be marked as 'secondary', and skip those here
-    return checkHealth(healthDependencies.values());
+    List<HealthDependency> primaryHealthDependencies = healthDependencies.values().stream()
+        .filter(HealthDependency::getPrimary)
+        .collect(Collectors.toList());
```
johnlcox: Is there a way to avoid converting this back into a list? The `checkHealth` method really just needs to be able to iterate over the dependencies and doesn't actually need it to be a list.

It's likely a very small list right now, so it probably doesn't matter for performance. If it's easy to do, then we should change this, but I'm not sure the best way while using streams.

cc: @nathanschile

health/service/src/main/java/com/cerner/beadledom/health/HealthDependency.java
johnlcox: Might be beneficial to add an alternative constructor that allows setting the primary value. An explicit default constructor would need to also be added for passivity.

This would allow implementors to just set the value in the constructor rather than overriding the `getPrimary` method.

Also, `getPrimary` isn't a great name. Could that method be deprecated and a new `isPrimary` method be added that returns the result of calling `getPrimary`? `isPrimary` could even be final to encourage using the new constructor instead of overriding `getPrimary`.

johnlcox: All our other code that uses `getPrimary` can also be changed to call `isPrimary`

nab0310: Added here: https://github.com/nab0310/beadledom/commit/6a0bd44aac6b9099379a8be557aeabc47089dafa

health/service/src/main/java/com/cerner/beadledom/health/internal/HealthChecker.java
nab0310: I'm not sure if there's a way we can do it and preserve the streaming. The basic tenants of streaming in java is that we filter a set of values by a condition and the collect them at the end. 

We might be able to do it by turning it into an iterator and then turning that into an iterable but that seems like a bad solution. 

`healthDependencies.values().stream()
        .filter(HealthDependency::getPrimary).iterator();`

Then,

`Iterable<HealthDependency> iterable = () -> primaryHealthDependencies;`

http://www.lambdafaq.org/how-can-i-turn-an-iterator-into-an-iterable/

johnlcox: I saw that post as well. That doesn't read as cleanly as what we have now.

I think the best approach would be to refactor `checkHealth` completely so that we could use a `map` function on the stream we already have here to dependency DTOs. However, I don't think it's worth the effort on this PR.

health/service/src/main/java/com/cerner/beadledom/health/HealthDependency.java
johnlcox: Making it final won't affect consumers at all since it's new, and if it's implemented by just calling `getPrimary` it won't break consumers from a behavior perspective either.

health/adr/default_dependencies_primary.md
```diff
@@ -0,0 +1,18 @@
+## **Title**: Dependencies Default to Primary
```
johnlcox: I want to check on @nathanschile's opinion. Is this the desired structure for ADRs or should there be one adr directory at the root of the project and then the filenames/titles could provide the additional context for which module they relate too?

health/service/src/main/java/com/cerner/beadledom/health/HealthDependency.java
nab0310: Oh, I see. I got `getPrimary` and `isPrimary` mixed up

health/adr/default_dependencies_primary.md
nathanschile: I think we should have the adr directory at the root. This would help with global decisions that may are not limited to a particular module.

health/service/src/main/java/com/cerner/beadledom/health/HealthDependency.java
nab0310: I made the new methods and created constructors here: https://github.com/cerner/beadledom/pull/123/commits/55cb09b555062c4ae1c016add5f38d260667fb36

health/adr/default_dependencies_primary.md
nab0310: I moved the adr directory to root, I wasn't sure what directory structure to have, so I just put the adr under the `adr/health` directory. Do we want something like that? Or I was thinking about breaking it up by concept, in this case, having a `adr/dependency` directory. Opinions?

Commit: https://github.com/cerner/beadledom/pull/123/commits/a44583a8072c278fdc538d566510516b2cbb8f23

@johnlcox @nathanschile

health/service/src/main/java/com/cerner/beadledom/health/HealthDependency.java
```diff
@@ -9,7 +9,22 @@
...
+  /**
+   * A constructor for Heath Dependencies that sets the dependency to be a primary one.
+   */
+  public HealthDependency() {
+    primary = true;
+  }
+
+  /**
+   * A constructor that allows the user to pass in if the dependency is primary or not.
+   * @param primary A boolean to indicate if the health dependency is primary or not.
```
johnlcox: Add a line break before the at-params. 

The at-params should *not* be full sentences, so start with a lower case letter and do not include a period at the end.

health/service/src/main/java/com/cerner/beadledom/health/HealthDependency.java
```diff
@@ -35,10 +50,18 @@
   /**
    * Returns a Boolean to indicate if this check is a primary check for this service.
    */
+  @Deprecated
   public Boolean getPrimary() {
     return primary;
   }
 
+  /**
+   * Returns a Boolean to indicate if this dependency is a primary dependency for this service.
```
johnlcox: Even simpler for boolean methods is something like `Returns true if this is a primary dependency; otherwise false.`

nab0310: Changed the boolean methods here: https://github.com/cerner/beadledom/pull/123/commits/95686b23b05f48a497bd6fac0a15acfb9712165b

health/service/src/main/java/com/cerner/beadledom/health/HealthDependency.java
nab0310: Added the line break and made it not a sentence: https://github.com/cerner/beadledom/pull/123/commits/95686b23b05f48a497bd6fac0a15acfb9712165b

health/service/src/main/java/com/cerner/beadledom/health/HealthDependency.java
```diff
@@ -35,10 +50,18 @@
   /**
    * Returns a Boolean to indicate if this check is a primary check for this service.
```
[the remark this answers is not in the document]
nab0310: Added the deprecated tag as well as told them to use `{@link #isPrimary()}` to access and added a sentence about using the constructors instead of overriding the method: https://github.com/cerner/beadledom/pull/123/commits/95686b23b05f48a497bd6fac0a15acfb9712165b

PR #125, opened 2018-11-28
Swagger2 - Add empty multibinder for ModelConverter

venki09:

Fixing a defect with swagger2 module where no default implementation of ModelConverter was bound.

### How was it tested?
Consumed SNAPSHOT version in another service

### How to test

### DIFF
--- a/dependency-check-suppressions.xml
+++ b/dependency-check-suppressions.xml
@@ -16,4 +16,12 @@
     <gav regex="true">^org\.tukaani:xz:.*$</gav>
     <cve>CVE-2015-4035</cve>
   </suppress>
+  <!-- Vulnerability fix for guava is not released yet so we need to suppress it for the time being -->
+  <suppress>
+    <notes><![CDATA[
+   file name: auto-value-1.6.2.jar/META-INF/maven/com.google.guava/guava/pom.xml
+   ]]></notes>
+    <gav regex="true">^com\.google\.guava:guava:.*$</gav>
+    <cpe>cpe:/a:google:guava</cpe>
+  </suppress>
 </suppressions>
--- a/dependency-check-suppressions.xml
+++ b/dependency-check-suppressions.xml
@@ -21,7 +21,7 @@
     <notes><![CDATA[
    file name: auto-value-1.6.2.jar/META-INF/maven/com.google.guava/guava/pom.xml
    ]]></notes>
-    <gav regex="true">^com\.google\.guava:guava:.*$</gav>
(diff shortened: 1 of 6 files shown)


dependency-check-suppressions.xml
```diff
@@ -16,4 +16,12 @@
     <gav regex="true">^org\.tukaani:xz:.*$</gav>
     <cve>CVE-2015-4035</cve>
   </suppress>
+  <!-- Vulnerability fix for guava is not released yet so we need to suppress it for the time being -->
```
nathanschile: guava -> auto-value

swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerModule.java
```diff
@@ -62,20 +61,14 @@
...
  */
 public class SwaggerModule extends AbstractModule {
   @Override
   protected void configure() {
     requireBinding(SwaggerConfig.class);
 
     // Create empty multibinder in case no ModelConverter bindings exist
-    Multibinder<ModelConverter> swaggerModelConverterBinder = Multibinder.newSetBinder(
-        binder(), ModelConverter.class);
+    Multibinder.newSetBinder(binder(), ModelConverter.class);
```
nathanschile: How is this different than the code that exists, 
```
Multibinder<ModelConverter> swaggerModelConverterBinder = Multibinder.newSetBinder(
        binder(), ModelConverter.class);
```

dependency-check-suppressions.xml
johnlcox: Can we confirm, by directly depending on a bad guava version, that we still fail if we actually pull it in even with this suppression? We really only want to suppress the false-positive from the pom included in autovalue.

nab0310: When I manually force the guava version down to `23.4`, the build passes.

nab0310: It works, instead of 
`<gav regex="true">^com\.google\.guava:guava:.*$</gav>` 
we can specify a file path for it to look on: 
`<filePath regex="true">.*auto-value-1.6.2\.jar.*$</filePath>`

Resource: https://jeremylong.github.io/DependencyCheck/general/suppression.html

venki09: Awesome thanks! cfa713b

johnlcox: Now the comment just needs updated.

venki09: oops. Forgot to push the commit 3db9829

PR #137, opened 2019-05-02
Allow swagger modules to make api-docs request through http or https

nab0310:

If the service consuming beadledom was using swagger 2 and https, it would fail to load the api-docs of that service because it made the request over http. This allows the request to be make over https as well as some other QoL changes for the swagger 2 module. 

 [unpkg](https://unpkg.com/) is a CDN for everything on npm. The Swagger team recommends it in their [readme](https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/installation.md#unpkg) as a way to install the required javascript files. This is cleaner than loading the assets statically. 

### How was it tested?

Running `mvn clean install` and using the module in an external project and hitting it over https and making sure that the `meta/swagger/ui` page loaded successfully.

### How to test

### DIFF
--- a/swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerUiResource.java
+++ b/swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerUiResource.java
@@ -80,16 +80,16 @@ public class SwaggerUiResource {
    */
   private boolean isRequestSecure(SecurityContext securityContext, HttpHeaders httpHeaders) {
     return securityContext.isSecure()
-        || secureForwardedHeader(httpHeaders)
-        || secureXForwardedProtoHeader(httpHeaders);
+        || hasSecureForwardedHeader(httpHeaders)
+        || hasSecureXForwardedProtoHeader(httpHeaders);
   }
 
   /**
    * Checks the Forwarded header for a protocol value of https.
    * @param httpHeaders The {@link HttpHeaders} object containing the request headers.
-   * @return true if the `Forwarded` is present and contains https and false otherwise.
+   * @return true if the `Forwarded` header is present and contains https; false otherwise.
    */
-  private boolean secureForwardedHeader(HttpHeaders httpHeaders) {
+  private boolean hasSecureForwardedHeader(HttpHeaders httpHeaders) {
     String forwardedHeader = httpHeaders.getRequestHeaders().getFirst("Forwarded");
     Pattern forwardedPairs = Pattern.compile("proto=(?<protocolValue>[^;]*)(;|\\z)");
 
@@ -100,11 +100,10 @@ public class SwaggerUiResource {
   /**
    * Checks the X-Forwarded-Proto header for a value of https.
    * @param httpHeaders The {@link HttpHeaders} object containing the request headers.
-   * @return true if the `X-Forwarded-Proto` header exists and false otherwise.
+   * @return true if the `X-Forwarded-Proto` header is present and contains https; false otherwise.
    */
-  private boolean secureXForwardedProtoHeader(HttpHeaders httpHeaders) {
-    return httpHeaders.getRequestHeaders().getFirst("X-Forwarded-Proto") != null
-        && httpHeaders.getRequestHeaders().getFirst("X-Forwarded-Proto").equals("https");
+  private boolean hasSecureXForwardedProtoHeader(HttpHeaders httpHeaders) {
+    return "https".equals(httpHeaders.getRequestHeaders().getFirst("X-Forwarded-Proto"));
   }
 
 }
--- a/swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerUiResource.java
+++ b/swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerUiResource.java
@@ -1,14 +1,12 @@
 package com.cerner.beadledom.swagger;
 
+import com.cerner.beadledom.jaxrs.StreamingWriterOutput;
 import com.github.mustachejava.DefaultMustacheFactory;
 import com.github.mustachejava.MustacheFactory;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import java.io.InputStream;
-import java.io.OutputStreamWriter;
-import java.nio.charset.StandardCharsets;
 import java.util.Set;
-import java.util.function.Consumer;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
@@ -49,15 +47,11 @@ public class SwaggerUiResource {
   @Produces(MediaType.TEXT_HTML)
   public StreamingOutput getSwaggerUi(@Context UriInfo uriInfo, @Context SecurityContext securityContext) {
     String httpSchema = securityContext.isSecure() ? "https" : "http";
-    return output -> {
-      Consumer<OutputStreamWriter> consumer = outputStreamWriter -> MUSTACHE_FACTORY.compile("ui.mustache").execute(
-          outputStreamWriter, ImmutableMap.of(
-              "apiDocsUrl",
-              uriInfo.getBaseUriBuilder().scheme(httpSchema).path(SwaggerApiResource.class).build().toASCIIString()));
-      OutputStreamWriter writer = new OutputStreamWriter(output, StandardCharsets.UTF_8);
-      consumer.accept(writer);
-      writer.flush();
-    };
+    return StreamingWriterOutput.with(writer -> MUSTACHE_FACTORY.compile("ui.mustache").execute(
+        writer, ImmutableMap.of(
+            "apiDocsUrl",
+            uriInfo.getBaseUriBuilder().scheme(httpSchema).path(SwaggerApiResource.class).build().toASCIIString()))
+    );
   }
 
   // This is a terrible way to serve static assets, but this shouldn't see a lot of traffic.
--- a/swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerUiResource.java
+++ b/swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerUiResource.java
@@ -16,6 +16,7 @@ import javax.ws.rs.Produces;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
+import javax.ws.rs.core.SecurityContext;
 import javax.ws.rs.core.StreamingOutput;
 import javax.ws.rs.core.UriInfo;
 
@@ -46,12 +47,13 @@ public class SwaggerUiResource {
   @GET
   @Path("/ui")
   @Produces(MediaType.TEXT_HTML)
-  public StreamingOutput getSwaggerUi(@Context UriInfo uriInfo) {
+  public StreamingOutput getSwaggerUi(@Context UriInfo uriInfo, @Context SecurityContext securityContext) {
+    String httpSchema = securityContext.isSecure() ? "https" : "http";
     return output -> {
       Consumer<OutputStreamWriter> consumer = outputStreamWriter -> MUSTACHE_FACTORY.compile("ui.mustache").execute(
           outputStreamWriter, ImmutableMap.of(
               "apiDocsUrl",
-              uriInfo.getBaseUriBuilder().path(SwaggerApiResource.class).build().toASCIIString()));
+              uriInfo.getBaseUriBuilder().scheme(httpSchema).path(SwaggerApiResource.class).build().toASCIIString()));
       OutputStreamWriter writer = new OutputStreamWriter(output, StandardCharsets.UTF_8);
       consumer.accept(writer);
       writer.flush();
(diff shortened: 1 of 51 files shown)


swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerUiResource.java
```diff
@@ -44,11 +45,12 @@
   @GET
   @Path("/ui")
   @Produces(MediaType.TEXT_HTML)
-  public StreamingOutput getSwaggerUi(@Context UriInfo uriInfo) {
+  public StreamingOutput getSwaggerUi(@Context UriInfo uriInfo, @Context SecurityContext securityContext) {
+    String httpSchema = securityContext.isSecure() ? "https" : "http";
```
johnlcox: Will this work behind a reverse proxy that terminates HTTPS? I think this would depend on if the SecurityContext understands the `X-Forwarded-Proto` header or other equivalent headers.

johnlcox: I think `Schema` should be `Scheme`.

swagger2/src/main/resources/com/cerner/beadledom/swagger2/ui.mustache
```diff
@@ -4,9 +4,9 @@
   <head>
     <meta charset="UTF-8">
     <title>Swagger UI</title>
-    <link rel="stylesheet" type="text/css" href="./swagger-ui.css" >
-    <link rel="icon" type="image/png" href="./favicon-32x32.png" sizes="32x32" />
-    <link rel="icon" type="image/png" href="./favicon-16x16.png" sizes="16x16" />
+    <link rel="stylesheet" type="text/css" href="//unpkg.com/swagger-ui-dist@3/swagger-ui.css" >
```
johnlcox: Nice. Where did you learn about this CDN for the swagger assets?

nab0310: The swagger installation readme had a reference for it, its pretty neat: https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/installation.md#unpkg

nab0310: Also, if I'm reading it right the `@3` will update the minor version every time a new one comes out and I believe you can change it for increased granularity if you wanted to specify a minor version as well.

swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerUiResource.java
nab0310: `SecurityContext` doesn't understand those types of headers, but we can use `HttpHeaders` along with the `getRequestHeader()` method to determine if we are behind a reverse proxy. It might look a little something like this (without the tedious repetition of the common forms of the `X-Forwareded-Proto` header:
```java
public StreamingOutput getSwaggerUi(@Context UriInfo uriInfo,
      @Context SecurityContext securityContext, @Context HttpHeaders httpHeaders) {
    String httpScheme = isSecure() ? "https" : "http";
    return StreamingWriterOutput.with(writer -> MUSTACHE_FACTORY.compile("ui.mustache").execute(
        writer, ImmutableMap.of(
            "apiDocsUrl",
            uriInfo.getBaseUriBuilder().scheme(httpScheme).path(SwaggerApiResource.class).build().toASCIIString()))
    );
  }

  private Boolean isSecure(SecurityContext securityContext, HttpHeaders httpHeaders) {
    return securityContext.isSecure() ||
        !httpHeaders.getRequestHeader("X-Forwareded-Proto").isEmpty() ||
        !httpHeaders.getRequestHeader("Forwarded").isEmpty();
  }
```

nab0310: https://github.com/cerner/beadledom/pull/137/commits/6d4a428dd8851add16eafafe2d16fa87866c04bc

nab0310: This is what i came up with, https://github.com/cerner/beadledom/pull/137/commits/8257840cfaa020ebc07620a2dc70c7cfabecb91e

I'm struggling to come up with a good way to test it, I tried putting it in the beadledom-client and running it but I really haven't come up with a good way of mocking `@Context`.

swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerUiResource.java
```diff
@@ -66,4 +71,21 @@ public Response getSwaggerUiAsset(@PathParam("asset") String assetPath) {
...
     return Response.ok(stream).build();
   }
+
+  private boolean isRequestSecure(SecurityContext securityContext, HttpHeaders httpHeaders) {
+    return securityContext.isSecure()
+        || secureForwardedHeader(httpHeaders.getRequestHeaders().getFirst("Forwarded"))
+        || secureXForwardedProtoHeader(httpHeaders.getRequestHeaders().getFirst("X-Forwarded-Proto"));
+  }
+
+  private boolean secureForwardedHeader(String forwardedHeader) {
```
venki09: Might be worth including javadoc to explain what this is doing for better understanding

swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerUiResource.java
```diff
@@ -66,4 +71,21 @@ public Response getSwaggerUiAsset(@PathParam("asset") String assetPath) {
     InputStream stream = getClass().getResourceAsStream("ui-dist/" + assetPath);
     return Response.ok(stream).build();
   }
+
+  private boolean isRequestSecure(SecurityContext securityContext, HttpHeaders httpHeaders) {
```
johnlcox: A thought occurred to me. Is the correct scheme already on the `UriInfo` regardless of being behind a reverse proxy or not? I'm guessing the UriInfo would have the correct domain rather than some local domain that is part of a DCOS cluster, and if it has the correct domain it seems like it would have the correct scheme as well.

nab0310: From the UriInfo docs, `the URIs are relative to the base URI of the application`.  So I think it would contain the correct scheme (that the base application used) as well. 

Would there be any good way to test this besides dropping this in an application and making sure it runs?

https://docs.oracle.com/javaee/7/api/javax/ws/rs/core/UriInfo.html

johnlcox: Some Google searches make it seem like it might not be as easy as I was hoping. I think the only way to test will be to actually deploy a service behind a reverse proxy and try it.

nab0310: 😢

nab0310: Well I tested it with a service running locally under `https` and `http` and it worked in those two environments. https://github.com/cerner/beadledom/pull/137/commits/418167b1919359a438b0d8d96aab8aa55eb24e9f

nab0310:
I made a release candidate and discovered that using headers was the only way it works for the service that is available to me behind a reverse proxy. I added some javadoc and other changes to the header parsing method signatures here: https://github.com/cerner/beadledom/pull/137/commits/c0791045ce645ac3e79dee46eec7271b74bd1d34

swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerUiResource.java
```diff
@@ -66,4 +71,40 @@ public Response getSwaggerUiAsset(@PathParam("asset") String assetPath) {
...
+    return forwardedHeader != null
+        && forwardedPairs.matcher(forwardedHeader).group("protocolValue").contains("https");
+  }
+
+  /**
+   * Checks the X-Forwarded-Proto header for a value of https.
+   * @param httpHeaders The {@link HttpHeaders} object containing the request headers.
+   * @return true if the `X-Forwarded-Proto` header exists and false otherwise.
+   */
+  private boolean secureXForwardedProtoHeader(HttpHeaders httpHeaders) {
```
johnlcox: Would this be better named `containsSecureXForwardedProtoHeader` or `hasSecureXForwardedProtoHeader`?

swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerUiResource.java
```diff
@@ -66,4 +71,40 @@ public Response getSwaggerUiAsset(@PathParam("asset") String assetPath) {
...
+    return httpHeaders.getRequestHeaders().getFirst("X-Forwarded-Proto") != null
```
johnlcox: I think reversing the equals would remove the need to check for null:

```java
return "https".equals(httpHeaders.getRequestHeaders().getFirst("X-Forwarded-Proto"));
```

swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerUiResource.java
```diff
@@ -66,4 +71,40 @@ public Response getSwaggerUiAsset(@PathParam("asset") String assetPath) {
...
+    Pattern forwardedPairs = Pattern.compile("proto=(?<protocolValue>[^;]*)(;|\\z)");
+
+    return forwardedHeader != null
+        && forwardedPairs.matcher(forwardedHeader).group("protocolValue").contains("https");
+  }
+
+  /**
+   * Checks the X-Forwarded-Proto header for a value of https.
+   * @param httpHeaders The {@link HttpHeaders} object containing the request headers.
+   * @return true if the `X-Forwarded-Proto` header exists and false otherwise.
```
johnlcox: "@returns true if the `X-Forwarded-Proto` header exists and is equal to `https`; otherwise false."

swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerUiResource.java
```diff
@@ -66,4 +71,40 @@ public Response getSwaggerUiAsset(@PathParam("asset") String assetPath) {
...
+  private boolean isRequestSecure(SecurityContext securityContext, HttpHeaders httpHeaders) {
+    return securityContext.isSecure()
+        || secureForwardedHeader(httpHeaders)
+        || secureXForwardedProtoHeader(httpHeaders);
+  }
+
+  /**
+   * Checks the Forwarded header for a protocol value of https.
+   * @param httpHeaders The {@link HttpHeaders} object containing the request headers.
+   * @return true if the `Forwarded` is present and contains https and false otherwise.
```
johnlcox: "`Forwarded` is" -> "`Forwarded` header is"

swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerUiResource.java
johnlcox: Actually this could mostly match the doc from `secureForwardedHeader`.

swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerUiResource.java
```diff
@@ -66,4 +71,40 @@ public Response getSwaggerUiAsset(@PathParam("asset") String assetPath) {
...
+   */
+  private boolean secureForwardedHeader(HttpHeaders httpHeaders) {
```
johnlcox: Prefix with `has` or `contains`?

nab0310: Prefixed with `has`: https://github.com/cerner/beadledom/pull/137/commits/448ff0c922f16b3a5fed81879c72a7931826231e

swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerUiResource.java
nab0310: changed: https://github.com/cerner/beadledom/pull/137/commits/448ff0c922f16b3a5fed81879c72a7931826231e

swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerUiResource.java
nab0310: Changed to mirror the doc for `hasSecureForwardedHeader`: https://github.com/cerner/beadledom/pull/137/commits/448ff0c922f16b3a5fed81879c72a7931826231e

swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerUiResource.java
nab0310: Yep: https://github.com/cerner/beadledom/pull/137/commits/448ff0c922f16b3a5fed81879c72a7931826231e

swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerUiResource.java
nab0310: I changed it to `hasSecureXForwardedProtoHeader`.

nab0310:
Closing this until I get a standalone Request header filter made, will open a new PR and link it here when I am done.

nab0310:
New PR: https://github.com/cerner/beadledom/pull/149

PR #140, opened 2019-06-12
Add Jooq transaction hooks for executing actions on commit

johnlcox:

Adds a new transaction hooks class that can be used to register actions during a Jooq transaction to be executed after the transaction is successfully committed.

### How was it tested?

* New unit tests
* Used in a service and manually tested some transactions.

### How to test

### DIFF
--- a/docs/source/manual/jooq.rst
+++ b/docs/source/manual/jooq.rst
@@ -112,9 +112,78 @@ Below is an example of configuring the required binding uses a C3P0 pooling data
     }
   }
 
+Post commit actions
+~~~~~~~~~~~~~~~~~~~~
+
+Sometimes you need to perform an action only after the current transaction successfully commits. For example,
+updating or invalidating a cache.
+
+The JooqTransactionalHooks_ class provides the :code:`whenCommitted` method for registering callback actions that should be executed after a transaction is successfully committed. Any no-arguments lambda expression or :code:`Runnable` implementation can be passed as the action.
+
+.. code-block:: java
+
+  hooks.whenCommitted(() -> cache.invalidate())
+
+The action you register will be called immediately after the current transaction is successfully committed.
+
+If you call :code:`whenCommitted` while there isn't an active transaction, the action will be executed immediately.
+
+If the current transaction is rolled back instead of committed, then your action will be discarded and never called.
+
+Nested Transactions
+"""""""""""""""""""
+
+Nested transactions (savepoints) are handled correctly. Actions registered via :code:`whenCommitted` in a nested transaction will only be called after the outermost transaction is committed, and will not be called if a rollback to any outer savepoint occurs during the transaction.
+
+.. code-block:: java
+
+  @JooqTransactional
+  public void outer() {
+    hooks.whenCommitted(() -> foo());
+    inner();
+  }
+
+  @JooqTransactional {
+  public void inner {
+    hooks.WhenCommitted(() -> bar());
+  }
+
+  // foo() and then bar() will be called when the outer transaction is committed.
+
+.. code-block:: java
+
+  @JooqTransactional
+  public void outer() {
+    hooks.whenCommitted(() -> foo());
+    try {
+      inner();
+    } catch(Exception e) {
+      // Log or handle exception, but since we caught the inner exception the outer transaction will not be rolled back.
+    }
+  }
+
+  @JooqTransactional {
+  public void inner {
+    hooks.WhenCommitted(() -> bar());
+    throw new RuntimeException();  // Raising an exception will cause this inner transaction to rollback.
+  }
+
+  // foo() will be called, but not bar()
+
+Order of Execution
+""""""""""""""""""
+
+Actions registered with :code:`whenCommitted` for a given transaction are executed in the ordered they were registered.
+
+Exception Handling
+""""""""""""""""""
+
+If an action registered with :code:`whenCommitted` for a given transactions throws an exception, then no later registered actions in the same transaction will be called. Since the callback actions are executed *after* a successful commit, an exception in a callback will not cause the transaction to roll back.
+
 .. _DataSource: https://docs.oracle.com/javase/8/docs/api/javax/sql/DataSource.html
 .. _DSLContext: https://www.jooq.org/javadoc/latest/org/jooq/DSLContext.html
 .. _DSLContextProvider: https://github.com/cerner/beadledom/tree/master/jooq/src/main/java/com/cerner/beadledom/jooq/DSLContextProvider.java
 .. _JooqTransactional: https://github.com/cerner/beadledom/tree/master/jooq/src/main/java/com/cerner/beadledom/jooq/JooqTransactional.java
 .. _SQLDialect: https://www.jooq.org/javadoc/latest/org/jooq/SQLDialect.html
 .. _ThreadLocalJooqModule: https://github.com/cerner/beadledom/tree/master/jooq/src/main/java/com/cerner/beadledom/jooq/ThreadLocalJooqModule.java
+.. _JooqTransactionalHooks:https://github.com/cerner/beadledom/tree/master/jooq/src/main/java/com/cerner/beadledom/jooq/JooqTransactionalHooqs.java
--- a/docs/source/manual/jooq.rst
+++ b/docs/source/manual/jooq.rst
@@ -145,7 +145,7 @@ Nested transactions (savepoints) are handled correctly. Actions registered via :
 
   @JooqTransactional {
   public void inner {
-    hooks.WhenCommitted(() -> bar());
+    hooks.whenCommitted(() -> bar());
   }
 
   // foo() and then bar() will be called when the outer transaction is committed.
@@ -164,7 +164,7 @@ Nested transactions (savepoints) are handled correctly. Actions registered via :
 
   @JooqTransactional {
   public void inner {
-    hooks.WhenCommitted(() -> bar());
+    hooks.whenCommitted(() -> bar());
     throw new RuntimeException();  // Raising an exception will cause this inner transaction to rollback.
   }
 
--- a/docs/source/manual/jooq.rst
+++ b/docs/source/manual/jooq.rst
@@ -180,6 +180,13 @@ Exception Handling
 
 If an action registered with :code:`whenCommitted` for a given transactions throws an exception, then no later registered actions in the same transaction will be called. Since the callback actions are executed *after* a successful commit, an exception in a callback will not cause the transaction to roll back.
 
+Why no rollback hooks?
+""""""""""""""""""""""
+
+Rollback hooks are more difficult to implement robustly. A variety of things can cause an implicit rollback. For instance, if your process is killed, in the middle of a transaction, without a graceful shutdown, your rollback hooks would not run.
+
+Instead of using a rollback hook, try to invert what you are trying to do. Instead of trying to undo something when a transaction fails, instead use a commit hook to delay doing something until after the transaction succeeds.
+
 .. _DataSource: https://docs.oracle.com/javase/8/docs/api/javax/sql/DataSource.html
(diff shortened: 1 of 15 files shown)


docs/source/manual/jooq.rst
```diff
@@ -112,9 +112,78 @@ Below is an example of configuring the required binding uses a C3P0 pooling data
     }
   }
```
nathanschile: Add a section why you aren't implementing rollback hooks.

jooq/src/test/scala/com/cerner/beadledom/jooq/DSLContextProviderImplSpec.scala
```diff
@@ -82,16 +85,21 @@ class DSLContextProviderImplSpec extends
           }
         }
 
+        val threads = mutable.ListBuffer[Thread]()
```
nathanschile: Why was L88-L95 change required?

docs/source/manual/jooq.rst
```diff
@@ -112,9 +112,78 @@ Below is an example of configuring the required binding uses a C3P0 pooling data
...
+
+  @JooqTransactional
+  public void outer() {
+    hooks.whenCommitted(() -> foo());
+    inner();
+  }
+
+  @JooqTransactional {
+  public void inner {
+    hooks.WhenCommitted(() -> bar());
```
nab0310: `hooks.WhenCommitted` -> `hooks.whenCommitted`

AnthonyRoss:
I'm getting
```
[ERROR] ----------------------------------------------------
[ERROR] .NET Assembly Analyzer could not be initialized and at least one 'exe' or 'dll' was scanned. The 'mono' executable could not be found on the path; either disable the Assembly Analyzer or configure the path mono. On some systems mono-runtime and mono-devel need to be installed.
[ERROR] ----------------------------------------------------
```
during mvn clean install, I'm running again to double check

johnlcox:
@AnthonyRoss I've been having a lot of trouble with the dependency-check plugin on the Cerner network. It fails to download CVEs quite frequently and then fails the whole build. This never seems to happen in travis and eventually I can get it to build locally, but I'm not clear on what I'm doing locally that eventually succeeds.

docs/source/manual/jooq.rst
johnlcox: https://github.com/cerner/beadledom/pull/140/commits/9beddb37c97e7dc2a05cddd935a8738d5c541066

AnthonyRoss:
@johnlcox Finally got the build to work

jooq/src/test/scala/com/cerner/beadledom/jooq/DSLContextProviderImplSpec.scala
johnlcox: Previously it started the threads and then as soon as it reached the closing `}` of the it-block it would assume the test passed even if not all the threads had finished. 

This new way waits until all 5 threads are finished or fails the test if it takes longer than 5 seconds.

docs/source/manual/jooq.rst
johnlcox: https://github.com/cerner/beadledom/pull/140/commits/77d726f9e2db85a48ab8c615eabd5cb20c3e7429

docs/source/manual/jooq.rst
```diff
@@ -112,9 +112,85 @@ Below is an example of configuring the required binding uses a C3P0 pooling data
...
+The action you register will be called immediately after the current transaction is successfully committed.
+
+If you call :code:`whenCommitted` while there isn't an active transaction, the action will be executed immediately.
+
+If the current transaction is rolled back instead of committed, then your action will be discarded and never called.
+
+Nested Transactions
+"""""""""""""""""""
+
+Nested transactions (savepoints) are handled correctly. Actions registered via :code:`whenCommitted` in a nested transaction will only be called after the outermost transaction is committed, and will not be called if a rollback to any outer savepoint occurs during the transaction.
```
nathanschile: I think that `Nested transactions (savepoints) are handled correctly. ` can be removed.

jooq/src/main/java/com/cerner/beadledom/jooq/JooqTransactionalHooks.java
```diff
@@ -0,0 +1,55 @@
...
+   * If a callback action throws an exception, no later registered callbacks in the same
+   * transaction will be executed.
+   * </p>
+   *
+   * @param action the action to be executed upon the successful commit of the current Jooq
+   *     transaction
+   */
+  public abstract void whenCommitted(Runnable action);
+
+  void setTransaction(JooqTransaction transaction) {
```
charanpanchagnula: Perhaps this is subjective, but i always think twice before having multiple implementations within an inheritance hierarchy like you have for the setter/getter  here and in the `ThreadLocalJooqTransactionalHooks` class. Why do we need these methods in two places?

Also, it seems like this could be avoided if we make this an interface and have the setter/getter as default methods because this abstract class does not have state associated to it.

jooq/src/main/java/com/cerner/beadledom/jooq/ThreadLocalCommitHookExecutingTransactionListener.java
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * @author John Leacox
+ * @since 3.3
+ */
+class ThreadLocalCommitHookExecutingTransactionListener extends DefaultTransactionListener {
+  private final ThreadLocal<Deque<JooqTransaction>> transactionDeques = new ThreadLocal<>();
+
+  // This assumes that the hooks implementation is using ThreadLocal or some other means to have
+  // the same lifecycle as this listener.
+  private final JooqTransactionalHooks transactionalHooks;
```
charanpanchagnula: Why would this type be anything other than ThreadLocalJooqTransactionalHooks? Maybe my question shines a light on the fact that I don't completely understand this disclaimer

docs/source/manual/jooq.rst
```diff
@@ -112,9 +112,85 @@ Below is an example of configuring the required binding uses a C3P0 pooling data
     }
   }
 
+Post commit actions
+~~~~~~~~~~~~~~~~~~~~
+
+Sometimes you need to perform an action only after the current transaction successfully commits. For example,
+updating or invalidating a cache.
+
+The JooqTransactionalHooks_ class provides the :code:`whenCommitted` method for registering callback actions that should be executed after a transaction is successfully committed. Any no-arguments lambda expression or :code:`Runnable` implementation can be passed as the action.
```
charanpanchagnula: Nit: There is an underscore after `JooqTransactionalHooks`

johnlcox: It's a link down to below: https://github.com/cerner/beadledom/pull/140/files#diff-2a3927f61725d9baa48fb9c016e79882R196

jooq/src/main/java/com/cerner/beadledom/jooq/JooqTransactionalHooks.java
johnlcox: The goal of how I split this up was to have a public facing API with only the `whenCommitted` method and then internal-only APIs necessary for the framework.

I think I tried this once before, but I want to try it again: what might work is to make this an interface with only `whenCommitted`, move `setTransaction` and `clearTransaction` to the implementation only, and Update `ThreadLocalCommitHookExecutingTransactionListener` to depend directly on the implementation. If we had multiple implementations we might need a different structure, but with only one implementation currently, if this works I think it should be fine.

johnlcox: Ahh, one disadvantage of this is that we'll need a Guice binding on the implementation which exposes it beyond just this package.

johnlcox: Looks like we already bind some package-private classes in Guice.

charanpanchagnula: I am always wary of any tool that uses reflections so heavily (Heard some horror stories about powermock)

johnlcox: Changed to an interface: https://github.com/cerner/beadledom/pull/140/commits/ed118b69f588ab7f1fe9de2e975b4236cdd2e7a2

jooq/src/main/java/com/cerner/beadledom/jooq/ThreadLocalCommitHookExecutingTransactionListener.java
johnlcox: I removed this comment with the change to an interface: https://github.com/cerner/beadledom/pull/140/commits/ed118b69f588ab7f1fe9de2e975b4236cdd2e7a2

docs/source/manual/jooq.rst
johnlcox: https://github.com/cerner/beadledom/pull/140/commits/9fb82820a6b7e1c326a58907464a9fdacce4eb1f

docs/source/manual/jooq.rst
nathanschile: nit. you have an `_` appended to on `JooqTransactionalHooks_`

johnlcox: It's intentional: https://github.com/cerner/beadledom/pull/140#discussion_r297306196

nathanschile: Sorry I missed the previous comment. The link has the trailing _ when I view the page in github, https://github.com/cerner/beadledom/blob/9fb82820a6b7e1c326a58907464a9fdacce4eb1f/docs/source/manual/jooq.rst while the other links don't.

docs/source/manual/jooq.rst
```diff
@@ -112,9 +112,85 @@ Below is an example of configuring the required binding uses a C3P0 pooling data
...
+
+Instead of using a rollback hook, try to invert what you are trying to do. Instead of trying to undo something when a transaction fails, instead use a commit hook to delay doing something until after the transaction succeeds.
+
 .. _DataSource: https://docs.oracle.com/javase/8/docs/api/javax/sql/DataSource.html
 .. _DSLContext: https://www.jooq.org/javadoc/latest/org/jooq/DSLContext.html
 .. _DSLContextProvider: https://github.com/cerner/beadledom/tree/master/jooq/src/main/java/com/cerner/beadledom/jooq/DSLContextProvider.java
 .. _JooqTransactional: https://github.com/cerner/beadledom/tree/master/jooq/src/main/java/com/cerner/beadledom/jooq/JooqTransactional.java
 .. _SQLDialect: https://www.jooq.org/javadoc/latest/org/jooq/SQLDialect.html
 .. _ThreadLocalJooqModule: https://github.com/cerner/beadledom/tree/master/jooq/src/main/java/com/cerner/beadledom/jooq/ThreadLocalJooqModule.java
+.. _JooqTransactionalHooks:https://github.com/cerner/beadledom/tree/master/jooq/src/main/java/com/cerner/beadledom/jooq/JooqTransactionalHooqs.java
```
nathanschile: Do you need a space after the first : ?

docs/source/manual/jooq.rst
johnlcox: I could be mistaken, but I believe it doesn't render correctly since the file being linked to doesn't actually exist in master yet.

docs/source/manual/jooq.rst
johnlcox: Probably, maybe that is why the underscore is rendering weird rather than what I mentioned above.

johnlcox: Yes that fixed the link rendering: https://github.com/cerner/beadledom/pull/140/commits/03a822b2d469696fbcb6e89a78fd519211748b5f

Issue #141, opened 2019-06-18
Wrong ObjectReader Being Used When Beadledom Clients Are Injected Before ResteasyModule


lulzWill:
#### Specs
Device: Macbook Pro
OS version: High Sierra 10.13.3
Java SDK version: 1.8.0_192-b12
Beadledom version: 3.2.5

#### The Problem:
Currently when using the ResteasyModule alongside other modules for Beadledom Service Clients, the installation order of the modules will determine which ObjectReader is used to deserialize a request body. I believe that this is caused by the way that the ResteasyProviderFactory is generated within [ResteasyContextListener](https://github.com/cerner/beadledom/blob/master/resteasy/src/main/java/com/cerner/beadledom/resteasy/ResteasyContextListener.java#L51). This will simply iterate through all of the bindings from our injector and then add all of the relevant/available providers to the [ResteasyProviderFactory](https://github.com/resteasy/Resteasy/blob/3.6.3.Final/resteasy-guice/src/main/java/org/jboss/resteasy/plugins/guice/ModuleProcessor.java#L45-L49) for use later. This causes issues when you are using the AnnotatedJacksonModule since it will lead to there being multiple providers bound within the injector. Resteasy will then simply use the first one that was bound during injection to handle requests. This can lead to ObjectMappers that were annotated to only be used within a service client to be used for the entire service instead.
#### Steps to reproduce:

  1. Find or Create a beadledom service with a POST endpoint that uses Avro to deserialize the post body(Uses AvroJacksonGuiceModule) that also consumes a beadledom service client.
  2. Install the beadledom service client module before installing the ResteasyModule/AvroJacksonGuiceModule.
  3. Start up the service and make a POST request to that service endpoint with an empty JSON object as the post body, ie. `{}`
  4. Service will fail to set the default values for the avro model.

#### Observed Results:

  * Service fails to set default values on avro post body since wrong ObjectReader is being used by Resteasy during the request.

#### Expected Results:

  * Expect annotated JacksonJsonProviders to not have any effect on the service's JacksonJsonProvider

PR #143, opened 2019-07-02
Exclude JAX-RS providers with binding annotations to be registered in the resteasy provider factory for the server

charanpanchagnula:
Changes are for addressing https://github.com/cerner/beadledom/issues/141

### How was it tested?
White box tests were written to test this change.

### How to test


charanpanchagnula:
My knowledge on this repo is woefully inadequate for me to decipher the existing integration tests so i haven't written any. If the changes warrant some, i'll put some effort and get myself up to speed with the existing integration tests.

resteasy/src/main/java/com/cerner/beadledom/resteasy/ModuleProcessor.java
```diff
@@ -0,0 +1,80 @@
...
+  private final ResteasyProviderFactory providerFactory;
+
+  /**
+   * Constructor for ModuleProcessor.
+   *
+   * @param registry {@link Registry} for adding resource implementation endpoints
+   * @param providerFactory {@link ResteasyProviderFactory} used for registering JAX-RS {@link Provider} instances
+   */
+  public ModuleProcessor(final Registry registry, final ResteasyProviderFactory providerFactory) {
+    Objects.requireNonNull(registry, "registry: null");
```
nab0310: What exception does this throw if registry is null? Is it a NullPointer? Would it make more sense as an ArgumentException?

charanpanchagnula: Using IllegalArgumentException vs NullPointerException is a holy war i don't want to start now. If i want to use IAE, i'll have to use Preconditions.checkArgument which isn't a native utility so i stuck with Objects, which is also something used throughout beadledom

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@
 * Update maven archetype for project generation with new client recommendations.
 * Update client documentation to recommend using an abstracted resource pattern.
 
+### Defects Corrected
+* Exclude JAX-RS providers with binding annotations to be registered in the resteasy provider factory for the server.
```
johnlcox: With the 3.3 release we are switching to the Keep a Changelog style for the Changelog - https://keepachangelog.com/en/1.0.0/

This section would become `Fixed`. The sentence structure may change a tiny bit too, there are some examples on the link above.

resteasy/src/main/java/com/cerner/beadledom/resteasy/ResteasyContextListener.java
```diff
@@ -3,14 +3,14 @@
 import com.cerner.beadledom.lifecycle.GuiceLifecycleContainers;
 import com.cerner.beadledom.lifecycle.LifecycleContainer;
 import com.cerner.beadledom.lifecycle.LifecycleInjector;
-import com.google.common.collect.Lists;
 import com.google.inject.Injector;
 import com.google.inject.Module;
+
```
johnlcox: There shouldn't be any line breaks for regular imports with the Google style guide.

charanpanchagnula: Could you provide a link to the style guide?

resteasy/src/main/java/com/cerner/beadledom/resteasy/ModuleProcessor.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Processes a given life cycle injector that contains Guice modules.
+ *
+ * <p>This class expands upon {@link org.jboss.resteasy.plugins.guice.ModuleProcessor} by enforcing
+ * restrictions on available JAX-RS {@link Provider}s to be added to {@link ResteasyProviderFactory}.
+ *
+ * @author Charan Panchagnula
+ */
+public class ModuleProcessor {
```
johnlcox: Can this be package private?

resteasy/src/main/java/com/cerner/beadledom/resteasy/ResteasyContextListener.java
```diff
@@ -41,7 +41,7 @@ public void contextInitialized(ServletContextEvent event) {
 
     final List<? extends Module> appModules = getModules(context);
 
-    List<Module> modules = Lists.newArrayList();
+    final List<Module> modules = new ArrayList<>();
```
johnlcox: We shouldn't be making local variables final.

resteasy/src/main/java/com/cerner/beadledom/resteasy/ResteasyContextListener.java
johnlcox: https://google.github.io/styleguide/javaguide.html

resteasy/src/main/java/com/cerner/beadledom/resteasy/ResteasyContextListener.java
charanpanchagnula: Why not? It allows compiler to make some optimizations and makes a great case for the kind of defensive coding we should all indulge in. Frankly i find the lack of final keyword in the class signature, method signatures and transient variables not only a stark departure from my previous team, but also a little too tolerant

resteasy/src/main/java/com/cerner/beadledom/resteasy/ModuleProcessor.java
```diff
@@ -0,0 +1,80 @@
...
+   */
+  public void processInjector(final Injector injector) {
+    Objects.requireNonNull(injector, "injector: null");
+
+    final List<Binding<?>> rootResourceBindings = new ArrayList<>();
+    for (final Binding<?> binding : injector.getBindings().values()) {
+      final Type type = binding.getKey().getTypeLiteral().getRawType();
+      final Class<?> beanClass = (Class) type;
+      if (GetRestful.isRootResource(beanClass)) {
+        // deferred registration
```
b-boogaard: Comments, especially non-javadoc ones like this, should really explain why not what. Right now this comment is more confusing than helpful to me.

charanpanchagnula: This was basically copy pastad from https://github.com/resteasy/Resteasy/blob/master/resteasy-guice/src/main/java/org/jboss/resteasy/plugins/guice/ModuleProcessor.java#L42 I can remove it

resteasy/src/test/java/com/cerner/beadledom/resteasy/TestProvider.java
```diff
@@ -0,0 +1,20 @@
...
+
+import javax.inject.Provider;
+
+/**
+ * Provides a fully-constructed instance of {@link TestJaxRsProvider}.
+ */
+public class TestProvider implements Provider<TestJaxRsProvider> {
+
+    /**
+     * Constructor for TestProvider.
```
b-boogaard: This comment isn't helpful.

resteasy/src/test/java/com/cerner/beadledom/resteasy/TestProviderWithBindingAnnotation.java
```diff
@@ -0,0 +1,29 @@
...
+import javax.inject.Provider;
+import java.lang.annotation.Annotation;
+
+/**
+ * Provides a fully-constructed instance of {@link TestJaxRsProvider} by taking in a binding annotation.
+ */
+public class TestProviderWithBindingAnnotation implements Provider<TestJaxRsProvider> {
+
+  /**
+   * Constructor for TestProviderWithBindingAnnotation.
```
b-boogaard: This comment isn't helpful. If an author cannot tell that this is a constructor they shouldn't be messing with it.

resteasy/src/test/java/com/cerner/beadledom/resteasy/TestJaxRsProvider.java
```diff
@@ -0,0 +1,16 @@
+package com.cerner.beadledom.resteasy;
+
+import javax.ws.rs.ext.Provider;
+
+/**
+ * JAX-RS provider for testing.
```
b-boogaard: Why do you add your `@author` tag sometimes and not others?

resteasy/src/main/java/com/cerner/beadledom/resteasy/ResteasyContextListener.java
johnlcox: Marking local variables as `final` doesn't provide much value and mostly just adds cruft to the method. If a method is sufficiently long that there is concern about a local variable being re-assigned, then the method should be split up; adding `final` is really just a band-aid for a bigger problem. The same is true of method parameters. Do you have an example of what compiler optimizations could be done for final local variables?

There sometimes is value in adding the `final` keyword to classes, but usually this would be for actual data classes. We usually solve this by using immutable data classes via autovalue. There may be some cases where marking methods as `final` is valuable but I haven't seen many.

There is definitely value in marking member variables as `final` since it indicates that no operations on that class will be re-assigning the member variables. This also indicates there is no need to worry about threading concerns related to those variables.

Following of strict process is not the same thing as creating high quality code. Adding `final` everywhere could be a process, but in reality it doesn't provide value in terms of code readability, maintainability, or defensiveness, for things like local variables and method parameters.

One thing I've found helpful in measuring my own code quality is to be able to have examples of high quality code to compare against. In Java I find the code of many of the Square open source libraries to be very high standards of good quality (e.g. https://github.com/square/retrofit and https://github.com/square/okhttp). No code-base is going to be perfect, but I would rather shoot for high quality code on principles and reason rather than by following a strict process.

charanpanchagnula: But are all those things mutually exclusive? Adding `final` wherever we can should not make the code any less readable or maintainable, nor does it violate any principles. If it does, then there are problems with the design. If we design something that is clearly NOT intended for extensibility or mutability, why would/should we even provide that flexibility for consumers (even if that consumers of something are just us) to potentially abuse our API? Isn't the functioning of our software under unforeseen circumstances like this the definition of defensive coding? I feel like having that mindset for local variables and method parameters is where it starts. And i don't understand how locking down our code in whatever way we can contributes in chipping away at code quality.

The above paragraph is just me providing a passionate defense for a mindset that has served me well. But at the end of the day, I've seen teams being stringent to a fault and teams being indifferent in this area and the scales aren't tipped in any particular direction in terms of the results. I'll remove the keyword.

resteasy/src/main/java/com/cerner/beadledom/resteasy/ModuleProcessor.java
johnlcox: If this is just copy pasted, do we need to include a copyright notice for Resteasy?

resteasy/src/test/scala/com/cerner/beadledom/resteasy/ModuleProcessorSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+    it("throws a NullPointerException for null injector") {
+      intercept[NullPointerException] {
+        new ModuleProcessor(registry, providerFactory).processInjector(null)
+      }
+    }
+
+    it("excludes providers that have a binding annotation") {
+      val injector = Guice.createInjector(new AnnotatedModule(classOf[TestBindingAnnotation]) {
+        override def configure(): Unit = {
+          bind(classOf[TestJaxRsProvider]).annotatedWith(getBindingAnnotation).toProvider(
```
johnlcox: I don't think the call to `toProvider` is needed for this test. Using just `bind(...).annotatedWith(...)` should test the same thing.

resteasy/src/main/java/com/cerner/beadledom/resteasy/ResteasyContextListener.java
b-boogaard: >  why would/should we even provide that flexibility for consumers (even if that consumers of something are just us) to potentially abuse our API?

How does having a final variable inside a method protect that method?

charanpanchagnula: That statement wasn't for the final variable, just extending the argument to classes that do not have a final keyword

johnlcox: My previous statements mention that there is sometimes value in making a class final, but I'm not going to generalize it to say there is always value. That is one I would want to look at on a case-by-case basis.

resteasy/src/test/scala/com/cerner/beadledom/resteasy/ModuleProcessorSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+
+  describe("ModuleProcessor#processInjector") {
+    it("throws a NullPointerException for null injector") {
+      intercept[NullPointerException] {
+        new ModuleProcessor(registry, providerFactory).processInjector(null)
+      }
+    }
+
+    it("excludes providers that have a binding annotation") {
+      val injector = Guice.createInjector(new AnnotatedModule(classOf[TestBindingAnnotation]) {
```
johnlcox: I think this should just be an `AbstractModule`.

resteasy/src/test/scala/com/cerner/beadledom/resteasy/ModuleProcessorSpec.scala
johnlcox: You can do `.annotatedWith(classOf[TestBindingAnnotation])`.

resteasy/src/main/java/com/cerner/beadledom/resteasy/ResteasyContextListener.java
b-boogaard: I'm with @johnlcox I think `final` has its uses and places that it will provide value. I'm just not a fan in blanket applying rules (some call it process) without any critical thinking about what makes sense for the current context.

resteasy/src/test/scala/com/cerner/beadledom/resteasy/ModuleProcessorSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+          expose(classOf[TestJaxRsProvider]).annotatedWith(getBindingAnnotation)
+
+          bind(classOf[TestJaxRsProvider]).toProvider(new TestProvider())
+          expose(classOf[TestJaxRsProvider])
+        }
+      })
+      val providerFactory = new ResteasyProviderFactory()
+      new ModuleProcessor(registry, providerFactory).processInjector(injector)
+
+      providerFactory.getProviderInstances must have size 1
```
johnlcox: Would it be possible to use 2 different JAX-RS provider classes in the Guice module above, and then have an assertion on the particular class/instance that is expected here?

resteasy/src/main/java/com/cerner/beadledom/resteasy/ModuleProcessor.java
b-boogaard: Why are we just copy pasting code?

resteasy/src/test/scala/com/cerner/beadledom/resteasy/ModuleProcessorSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+import org.scalatest.mockito.MockitoSugar
+import com.cerner.beadledom.guice.dynamicbindings.AnnotatedModule
+import com.google.inject.Guice
+
+class ModuleProcessorSpec extends FunSpec with MockitoSugar with MustMatchers{
+
+  val registry: Registry = mock[Registry]
+  val providerFactory: ResteasyProviderFactory = mock[ResteasyProviderFactory]
+
+  describe("ModuleProcessor#new") {
```
johnlcox: I think in other tests we use nested `describe` blocks for the class methods.

```scala
describe("ModuleProcessor") {
  describe("constructor") {
  }

  describe("#processInjector") {
  }
}
```

resteasy/src/main/java/com/cerner/beadledom/resteasy/ResteasyContextListener.java
charanpanchagnula: Making a class or a method parameter final when it is created specifically to be extended or re-assigned..THAT would be applying rules without any critical thinking, and thats not what i am advocating. I am saying that applying the rules when you're really indifferent about the extensibility of a class or mutability of a variable is okay and following this rule doesn't mean that it is senseless for a current context.

Again that being said, i am happy to remove the keyword for the code in this PR

resteasy/src/main/java/com/cerner/beadledom/resteasy/ModuleProcessor.java
charanpanchagnula: Like i added in this comment, ModuleProcessor in resteasy-guice does not provide the functionality we want to address this issue so we decided to write our own version of this class https://github.com/charanpanchagnula/beadledom/blob/issue16_clientBindingAnnotation/resteasy/src/main/java/com/cerner/beadledom/resteasy/ModuleProcessor.java#L23

resteasy/pom.xml
```diff
@@ -98,6 +98,11 @@
       <artifactId>beadledom-swagger1</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.cerner.beadledom</groupId>
+      <artifactId>beadledom-guice-dynamicbindings</artifactId>
```
nathanschile: nit. This dependency should be above beadledom-swagger1.

johnlcox: It can be removed completely after some of the changes I suggested in the tests.

resteasy/src/main/java/com/cerner/beadledom/resteasy/ModuleProcessor.java
```diff
@@ -0,0 +1,80 @@
...
+        rootResourceBindings.add(binding);
+      }
+      if (beanClass.isAnnotationPresent(Provider.class)) {
+        // For a single running resteasy instance, there is no reason to register annotated bindings
+        // because it can potentially lead to multiple providers of the same type bound within the injector.
+        if (binding.getKey().getAnnotationType() == null) {
```
nathanschile: Do you think this makes sense for this change to be merged upstream in resteasy?

resteasy/src/main/java/com/cerner/beadledom/resteasy/ModuleProcessor.java
```diff
@@ -0,0 +1,80 @@
+package com.cerner.beadledom.resteasy;
+
+import com.google.inject.Binding;
```
nathanschile: Another option available would be to update `com.cerner.beadledom.jaxrs.provider.FilteringJacksonJsonProvider` class and add  the `@Priority(a number < 5000)` annotation. RESTEasy uses the priority to sort the providers [1]. FilteringJacksonJsonProvider is only used server-side so this would be a safe change to make.

[1] 
https://github.com/resteasy/Resteasy/blob/3.6/resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java#L2176-L2177
https://github.com/resteasy/Resteasy/blob/3.6/resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java#L1723-L1725

resteasy/src/main/java/com/cerner/beadledom/resteasy/ModuleProcessor.java
nathanschile: I'll answer my own question, probably not.

resteasy/src/main/java/com/cerner/beadledom/resteasy/ResteasyContextListener.java
charanpanchagnula: Also, why are local variables from 34-41 set to final? Shouldn't we remove the final keyword everywhere if we are this particular about enforcing it?

johnlcox: I'm not sure why the variables on lines 34-41 are set to final. It would be good to fix those too.

charanpanchagnula: Looks like it's because checkstyle complains when the variable is defined far enough from its usage.  `Consider making that variable final if you still need to store its value in advance` is the fix that checkstyle is advocating.

Even if we re-order the definition of these variables in way that doesn't sacrifice readability, checkstyle still finds it problematic. If the fix for this is to retain the final keywords, in my opinion it creates more confusion because we'll be stuck in this weird place where some of them are final and some of them aren't and readers will not know why. We could change checkstyle settings but i think we all agree that that wouldn't really be a solution.

johnlcox: It might be possible to rearrange and possibly split up the method, but if so it might be better to do as a separate PR.

resteasy/src/main/java/com/cerner/beadledom/resteasy/ModuleProcessor.java
```diff
@@ -0,0 +1,80 @@
...
+          LogMessages.LOGGER.info(Messages.MESSAGES.registeringProviderInstance(beanClass.getName()));
+          providerFactory.registerProviderInstance(binding.getProvider().get());
+        }
+      }
+    }
+    for (final Binding<?> binding : rootResourceBindings) {
+      final Class<?> beanClass = (Class) binding.getKey().getTypeLiteral().getType();
+      final ResourceFactory resourceFactory = new GuiceResourceFactory(binding.getProvider(), beanClass);
+
+      LogMessages.LOGGER.info(Messages.MESSAGES.registeringFactory(beanClass.getName()));
```
nathanschile: Should we be using the SLF4j Logger?

CHANGELOG.md
charanpanchagnula: https://github.com/cerner/beadledom/pull/143/commits/75d364f068be5b48f12a864aa2a8bd938fb929b8

resteasy/src/test/java/com/cerner/beadledom/resteasy/TestJaxRsProvider.java
charanpanchagnula: https://github.com/cerner/beadledom/pull/143/commits/75d364f068be5b48f12a864aa2a8bd938fb929b8

resteasy/src/test/java/com/cerner/beadledom/resteasy/TestProvider.java
charanpanchagnula: https://github.com/cerner/beadledom/pull/143/commits/75d364f068be5b48f12a864aa2a8bd938fb929b8

resteasy/src/test/java/com/cerner/beadledom/resteasy/TestProviderWithBindingAnnotation.java
charanpanchagnula: https://github.com/cerner/beadledom/pull/143/commits/75d364f068be5b48f12a864aa2a8bd938fb929b8

resteasy/src/test/scala/com/cerner/beadledom/resteasy/ModuleProcessorSpec.scala
charanpanchagnula: https://github.com/cerner/beadledom/pull/143/commits/75d364f068be5b48f12a864aa2a8bd938fb929b8

resteasy/src/test/scala/com/cerner/beadledom/resteasy/ModuleProcessorSpec.scala
charanpanchagnula: https://github.com/cerner/beadledom/pull/143/commits/75d364f068be5b48f12a864aa2a8bd938fb929b8

resteasy/src/test/scala/com/cerner/beadledom/resteasy/ModuleProcessorSpec.scala
charanpanchagnula: https://github.com/cerner/beadledom/pull/143/commits/75d364f068be5b48f12a864aa2a8bd938fb929b8

resteasy/pom.xml
charanpanchagnula: https://github.com/cerner/beadledom/pull/143/commits/75d364f068be5b48f12a864aa2a8bd938fb929b8

resteasy/src/test/scala/com/cerner/beadledom/resteasy/ModuleProcessorSpec.scala
charanpanchagnula: Cannot remove the toProvider because it throws an error.

`No implementation for com.cerner.beadledom.resteasy.TestJaxRsProvider annotated with interface com.cerner.beadledom.resteasy.TestBindingAnnotation was bound.
`

resteasy/src/main/java/com/cerner/beadledom/resteasy/ModuleProcessor.java
charanpanchagnula: https://github.com/cerner/beadledom/commit/75d364f068be5b48f12a864aa2a8bd938fb929b8

resteasy/src/main/java/com/cerner/beadledom/resteasy/ResteasyContextListener.java
charanpanchagnula: https://github.com/cerner/beadledom/commit/75d364f068be5b48f12a864aa2a8bd938fb929b8

resteasy/src/main/java/com/cerner/beadledom/resteasy/ResteasyContextListener.java
charanpanchagnula: Like i said, re-arranging isn't possible without making it less readable. I tried re-arranging it for this particular use case.

Sorry for continuing to stretch this conversation further, but I am not convinced at all that we have to split up the method to get around something that (in my opinion) isn't a problem at all. We're trying to twist ourselves into knots unnecessarily when this method is absolutely fine the way it is.

Anyway, i've removed the final keyword: https://github.com/cerner/beadledom/commit/75d364f068be5b48f12a864aa2a8bd938fb929b8

resteasy/src/main/java/com/cerner/beadledom/resteasy/ModuleProcessor.java
charanpanchagnula: I guess we're currently using slf4j logs transitively through resteasy anyway, so i am not sure i know the answer to this question

resteasy/src/test/scala/com/cerner/beadledom/resteasy/ModuleProcessorSpec.scala
```diff
@@ -0,0 +1,84 @@
...
+        val providerFactory = new ResteasyProviderFactory()
+        new ModuleProcessor(registry, providerFactory).processInjector(injector)
+
+        providerFactory.getProviderInstances mustBe empty
+      }
+
+      it("excludes provider with binding annotation when two providers of the same type are bound with the injector") {
+        val injector = Guice.createInjector(new AbstractModule {
+          override def configure(): Unit = {
+            bind(classOf[TestJaxRsProvider]).annotatedWith(classOf[TestBindingAnnotation]).toProvider(
```
johnlcox: The call to `toProvider` here and elsewhere aren't needed. Guice can create an instance of the binding via a constructor, and if there are multiple constructors it looks for the `@Inject` annotation to know which one to use.

resteasy/src/test/scala/com/cerner/beadledom/resteasy/ModuleProcessorSpec.scala
```diff
@@ -0,0 +1,84 @@
...
+      }
+
+      it("throws a NullPointerException for null provider factory") {
+        intercept[NullPointerException] {
+          new ModuleProcessor(registry, null)
+        }
+      }
+    }
+
+    describe("ModuleProcessor#processInjector") {
```
johnlcox: This can just be "#processInjector"; the output of scala-test will handle formatting based on the nesting.

resteasy/src/test/scala/com/cerner/beadledom/resteasy/ModuleProcessorSpec.scala
charanpanchagnula: https://github.com/cerner/beadledom/pull/143#discussion_r301580313 Are you suggesting i have to change the code in the TestJaxRxProvider class in some way?

johnlcox: No. I took a look at the Guice annotated binding docs [1] and it looks like you can't do it without some `#to` call. You should be able to do `.to(classOf[TestJaxRsProvider])` rather than having a custom Guice provider though.

[1] https://github.com/google/guice/wiki/BindingAnnotations

resteasy/src/test/scala/com/cerner/beadledom/resteasy/ModuleProcessorSpec.scala
charanpanchagnula: https://github.com/cerner/beadledom/pull/143/commits/baf1e83e23496f193739668eb9632289d1cfbc1c

resteasy/src/test/scala/com/cerner/beadledom/resteasy/ModuleProcessorSpec.scala
charanpanchagnula: https://github.com/cerner/beadledom/pull/143/commits/baf1e83e23496f193739668eb9632289d1cfbc1c

resteasy/src/main/java/com/cerner/beadledom/resteasy/ModuleProcessor.java
johnlcox: What about changes similar to this so we aren't copying a file from another project: https://gist.github.com/johnlcox/57c451dc51d010ffb6c139de67aaa3f7?

charanpanchagnula: Functionality is more or less the same, so why are we adding bells and whistles to make it seem like its our own thing, as opposed to acknowledging that we're building upon the resteasy version (which is already mentioned in the class doc) ?

charanpanchagnula: Also, your logic seems to altogether ignore annotated bindings whereas the current logic still adds the annotated binding to rootResourceBindings list. Don't we have to register resource factory for this annotated binding?

johnlcox: The functionality from Resteasy is fairly simple, and since we need to modify it in a way we cannot do via class extension I think it's simpler to write our own class/method rather than copying. 

If we are going to copy then we have a legal obligation to follow the requirements of the Apache License for the file we are copying. The StackExchange link below gives a good answer as to what we would need to do to correctly follow the license [1]. Of the steps outlined there steps 3 and 5 are going to be the ones requiring the most effort. For step 3 it will also result in muddying up the file if we have to document all of our own changes. Step 5 will require evaluating github repo of Resteasy to determine if there is a "NOTICE-like" file, and if making a decision on how/where we will keep it in our repo.



[1] https://opensource.stackexchange.com/questions/5143/how-to-mark-a-copied-apache-v2-piece-of-code

johnlcox: Whether we should exclude annotated bindings of root resources or not is a good question. I'm curious what @nathanschile thinks.

nathanschile: So root resources being classes that are annotated with `@Path` annotation, what would be the use case where those classes would be annotated binded? I can't think of a use case.

charanpanchagnula: https://github.com/cerner/beadledom/pull/143/commits/75cf28188e471c1af009d2dd0dfec137b1270635

there was an if-else if check for root resources and provider bindings which isn't originally there in the resteasy-guice version i stuck to if-if to align with the functionality of resteasy-guice

resteasy/src/main/java/com/cerner/beadledom/resteasy/InjectorProcessor.java
```diff
@@ -0,0 +1,98 @@
...
+   * likely related to HTTP client bindings. A service should only have one set of JAX-RS service
+   * bindings, so they should be bound in Guice without an annotation.</p>
+   *
+   * @param injector The {@link Injector} instance to be processed
+   */
+  public void process(final Injector injector) {
+    Objects.requireNonNull(injector, "injector: null");
+
+    Predicate<Binding<?>> isNonAnnotatedBinding = binding -> binding.getKey().getAnnotationType() == null;
+    List<Binding<?>> nonAnnotatedBindings = injector.getBindings().values().stream().filter(isNonAnnotatedBinding)
```
johnlcox: Nice. I like the use of streams for this.

resteasy/src/main/java/com/cerner/beadledom/resteasy/InjectorProcessor.java
```diff
@@ -0,0 +1,98 @@
...
+ * @since 3.3
+ */
+class InjectorProcessor {
+  private static final Logger logger = LoggerFactory.getLogger(InjectorProcessor.class);
+
+  private final Registry registry;
+  private final ResteasyProviderFactory providerFactory;
+
+  /**
+   * Creates a new instance for InjectorProcessor with thw given Resteasy resource registry
```
johnlcox: "thw" -> "the"

resteasy/src/test/scala/com/cerner/beadledom/resteasy/InjectorProcessorSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+      }
+
+      it("excludes provider with binding annotation when two providers of the same type are bound with the injector") {
+        val injector = Guice.createInjector(new AbstractModule {
+          override def configure(): Unit = {
+            bind(classOf[TestJaxRsProvider]).annotatedWith(classOf[TestBindingAnnotation]).to(
+              classOf[TestJaxRsProvider]
+            )
+
+            bind(classOf[TestJaxRsProvider]).toProvider(new TestProvider())
```
johnlcox: I think this line could be just `bind(classOf[TestJaxRsProvider])` and would still work. Then the `TestProvider` class wouldn't need to exist.

charanpanchagnula: https://github.com/cerner/beadledom/pull/143/commits/bd6ab90bbf3a2dfb902bc7cda220fbd0a1cae674

resteasy/src/main/java/com/cerner/beadledom/resteasy/InjectorProcessor.java
charanpanchagnula: https://github.com/cerner/beadledom/pull/143/commits/bd6ab90bbf3a2dfb902bc7cda220fbd0a1cae674

PR #146, opened 2019-07-17
Deprecate pagination list metadata convenience overloads

### LINKED ISSUE #130: Deprecate metadata methods in OffsetPaginatedList
The `metadata(...)` methods in OffsetPaginatedList seem clunky to use in consumers. The three `OffsetPaginatedList#metadata` methods should be deprecated. The metadata builder should be the preferred way of building metadata if you need it, instead of calling overloaded methods. The `OffsetPaginatedList#metadata(Long)` can documented to prefer a new `OffsetPaginatedList#totalResults(Long)` and the other two can be documented to just use the metadata builder.

johnlcox:
This fixes #130.

### DIFF
--- a/pagination/src/main/java/com/cerner/beadledom/pagination/OffsetPaginatedList.java
+++ b/pagination/src/main/java/com/cerner/beadledom/pagination/OffsetPaginatedList.java
@@ -32,10 +32,9 @@ public abstract class OffsetPaginatedList<T> {
    * @return instance of {@link OffsetPaginatedList.Builder}
    */
   public static <T> OffsetPaginatedList.Builder<T> builder() {
-    return new AutoValue_OffsetPaginatedList.Builder<T>().items(new ArrayList<>())
-        .metadata(
-            LimitParameter.getDefaultLimitFieldName(), null,
-            OffsetParameter.getDefaultOffsetFieldName(), null, null, null);
+    return new AutoValue_OffsetPaginatedList.Builder<T>()
+        .items(new ArrayList<>())
+        .metadata(OffsetPaginatedListMetadata.builder().build());
   }
 
   @AutoValue.Builder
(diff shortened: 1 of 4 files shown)


pagination/src/main/java/com/cerner/beadledom/pagination/OffsetPaginatedList.java
```diff
@@ -93,14 +98,27 @@
...
+     *
+     * @param totalResults the total results available
+     * @return the {@link Builder} being used for the {@link OffsetPaginatedList}
+     */
+    public OffsetPaginatedList.Builder<T> totalResults(@Nullable Long totalResults) {
       return metadata(
-          LimitParameter.getDefaultLimitFieldName(), null,
-          OffsetParameter.getDefaultOffsetFieldName(), null,
-          totalResults, null);
+          OffsetPaginatedListMetadata.builder()
```
charanpanchagnula: Possibly stupid question: Do we have to do argument validation for totalResults to ensure it doesn't have negative values?

johnlcox: That's probably a good idea. Does anyone else have any passivity concerns of me adding such a check?

nathanschile: No concerns from me.

johnlcox: Autovalue recommends doing validation in the `build` method rather than directly in the setters [1]. This makes it a little bit less fail-fast, but should still be fine.

[1] https://github.com/google/auto/blob/master/value/userguide/builders-howto.md#validate

johnlcox: https://github.com/cerner/beadledom/pull/146/commits/91732ff7059233b1f33f364e3e4b4409723a40ab

PR #148, opened 2019-07-24
Update archetype to newer versions of dependencies

bbaugher:

Bumped the version of dependencies in the archetype to later versions that are not vulnerable to security vulnerabilities. I also ran into an issue when I used the archetype in 3.2.5 with `NoSuchMethodError` around guava (`java.lang.NoSuchMethodError: com.google.common.base.Preconditions.checkArgument(ZLjava/lang/String;Ljava/lang/Object;Ljava/lang/Object;)`) and after updating to a newer version (`java.lang.ClassNotFoundException: com.google.inject.multibindings.Multibinder`) since guice added multibinding extensions in a separate jar.

This also matches what beadledom libraries use

### How was it tested?

Create a sample project and updated the pom to get things working.

### How to test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Jooq transaction hooks for executing actions after a successful commit.
 
 ### Changed
+* Update archetype dependencies to match beadledom library
 * Switch to Keep a Changelog style for Changelog moving forward - https://keepachangelog.com/en/1.0.0/
 * Update maven archetype for project generation with new client recommendations.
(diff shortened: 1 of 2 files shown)


johnlcox:
Can you add an entry to the changelog? We are following https://keepachangelog.com/en/1.0.0/.

bbaugher:
Added, https://github.com/cerner/beadledom/pull/148/commits/5c3aa6fd967643c03371ed10e266cff85b4cdc77

Issue #131, opened 2018-12-13
Add https support to swaggerUI


venki09:
We need to have swagger UI support https protocol.

venki09:
https://github.com/cerner/beadledom/pull/137 was created to add this support

PR #149, opened 2019-07-29
Add ContainerRequestFilter to properly support services running behind proxies

nab0310:
If the service consuming beadledom was using swagger 2 and https behind a reverse-proxy or load balancer, it would fail to load the api-docs of that service because it made the request over http. This PR introduces a `ContainerRequestFilter` that reads the request context and determines if it originated from a secure context. 

An integration module was also added to test these changes. The module uses a simple service and a reverse proxy to properly test the changes.

This PR is an update to https://github.com/cerner/beadledom/pull/137 and references the issue https://github.com/cerner/beadledom/issues/131

### How was it tested?
`mvn clean install`

The new integration module directly tests the added functionality.

### How to test


integration/client/src/main/java/com/cerner/beadledom/integration/client/BeadledomIntegrationClientModule.java
```diff
@@ -0,0 +1,49 @@
...
+import com.google.inject.PrivateModule;
+import com.google.inject.Provides;
+import javax.inject.Singleton;
+
+/**
+ * An example Client Module.
+ *
+ * @author Nick Behrens
+ */
+public class BeadledomIntegrationClientModule extends PrivateModule {
```
johnlcox: This should probably be changed to follow our updated guidance. There could be an `IntegrationClient` class that has a method `#helloWorld` or `#helloWorldResource` rather than directly exposing the client JAX-RS resource.

johnlcox:
Add an entry to the changelog.

integration/client/src/main/java/com/cerner/beadledom/integration/client/BeadledomIntegrationClientModule.java
nab0310: Added a `BeadledomIntegrationClient` class: https://github.com/cerner/beadledom/pull/149/commits/0bb084b91c13cd5b595223d33244b4bb9b9d1421

nab0310:
> Add an entry to the changelog.

Added: https://github.com/cerner/beadledom/pull/149/commits/0bb084b91c13cd5b595223d33244b4bb9b9d1421

I'm not too solid on the actual sentence surrounding the `ForwardedHeaderFilter` so some feedback would be appreciated there.

integration/pom.xml
```diff
@@ -0,0 +1,108 @@
...
+      </plugin>
+    </plugins>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <artifactId>maven-war-plugin</artifactId>
+          <version>3.0.0</version>
+        </plugin>
+        <plugin>
+          <groupId>org.apache.tomcat.maven</groupId>
```
venki09: Is tomcat being used?

integration/service/src/main/java/com/cerner/beadledom/integration/service/ResteasyBootstrapModule.java
```diff
@@ -0,0 +1,54 @@
...
+import com.cerner.beadledom.resteasy.ResteasyModule;
+import com.cerner.beadledom.swagger2.Swagger2Module;
+import com.google.inject.AbstractModule;
+import com.google.inject.Provides;
+import com.google.inject.multibindings.OptionalBinder;
+import io.swagger.models.Info;
+import javax.inject.Singleton;
+
+/**
+ * A Guice Module that installs the {@link ResteasyModule}, {@link Swagger2Module},
```
venki09: nit: might be better to use List tag to list them

integration/service/src/main/java/com/cerner/beadledom/integration/service/resource/HelloWorldResource.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.ws.rs.core.MediaType;
+
+/**
+ * Server implementation of an offset based paginated list endpoint.
+ *
+ * @author Nick Behrens
+ */
+@Api(value = "/hello", description = "Retrieve hello world data")
+@Path("/hello")
+public class HelloWorldResource {
```
venki09: I think we can just extend the resource class from the api project?

integration/test/pom.xml
```diff
@@ -0,0 +1,205 @@
...
+
+  <dependencies>
+    <dependency>
+      <groupId>com.cerner.beadledom</groupId>
+      <artifactId>beadledom-integration-client</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.cerner.beadledom</groupId>
+      <artifactId>beadledom-integration-api</artifactId>
+    </dependency>
```
venki09: super nit: order this so that api comes first before client dependency

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/provider/ForwardedHeaderFilterSpec.scala
```diff
@@ -0,0 +1,266 @@
+package com.cerner.beadledom.jaxrs.provider
```
venki09: nit: describe and it blocks needs space after `{`

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/provider/ForwardedHeaderFilterSpec.scala
```diff
@@ -0,0 +1,266 @@
...
+    Mockito.when(request.getUriInfo).thenReturn(uriInfo)
+    request
+  }
+
+  describe("For a request with just the X-Forwarded-Proto header"){
+    describe("when the header has a value of https"){
+      it("changes the request to https"){
+        val request = createContainerRequestContext
+        val securityContext = mock[SecurityContext]
+        Mockito.when(securityContext.isSecure).thenReturn(false)
```
venki09: Can this be imported so we don't have to do `Mockito.when`?

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/provider/ForwardedHeaderFilterSpec.scala
```diff
@@ -0,0 +1,266 @@
...
+      describe("if the protocol is not set to https"){
+        it("keeps the request http"){
+          val request = createContainerRequestContext
+          val securityContext = mock[SecurityContext]
+          Mockito.when(securityContext.isSecure).thenReturn(false)
+          Mockito.when(request.getSecurityContext).thenReturn(securityContext)
+          Mockito.when(request.getHeaderString("X-Forwarded-Proto")).thenReturn(null)
+          Mockito.when(request.getHeaderString("Forwarded")).thenReturn("proto=other")
+
+
```
venki09: nit: extra line can be removed

integration/client/src/main/java/com/cerner/beadledom/integration/client/IntegrationClientFeature.java
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * A default client feature annotation.
+ *
+ * @author Nick Behrens
+ */
+@BindingAnnotation
+@Target({ElementType.FIELD, ElementType.PARAMETER, ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface IntegrationClientFeature {
```
venki09: Can this be package-private?

integration/api/src/main/java/com/cerner/beadledom/integration/api/model/HelloWorldDto.java
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * Creates a new {@link HelloWorldDto} instance.
+   *
+   * @param name author to be associated with the message
+   * @param message the message to be returned
+   * @return {@link HelloWorldDto} encapsulating a name and message
+   */
+  @JsonCreator
+  public static HelloWorldDto create(@JsonProperty("name") String name, @JsonProperty("hello_world_message") String message) {
```
venki09: Is this line too long?

integration/service/src/main/java/com/cerner/beadledom/integration/service/resource/HelloWorldResource.java
nab0310: With the current way they are set up, we cannot. The api project implementation of `getHelloWorld` returns an OffsetPaginatedListHelloWorldDto which implements the `OffsetPaginatedListDto` whereas the service resource returns a type of `OffsetPaginatedList`. That being said, I might be missing something obvious with how to set it up.

integration/api/src/main/java/com/cerner/beadledom/integration/api/HelloWorldResource.java
```diff
@@ -0,0 +1,35 @@
...
+import javax.ws.rs.core.MediaType;
+
+/**
+ * Client interface of an offset based paginated list endpoint resource.
+ *
+ * @author Nick Behrens
+ */
+@Api(value = "/hello", description = "Retrieve hello world data")
+@Path("/hello")
+public interface HelloWorldResource {
```
johnlcox: Since this is only for the client should it go to the client module instead of api?

integration/api/src/main/java/com/cerner/beadledom/integration/api/HelloWorldResource.java
```diff
@@ -0,0 +1,35 @@
+package com.cerner.beadledom.integration.api;
+
+import com.cerner.beadledom.integration.api.model.HelloWorldDto;
+import com.cerner.beadledom.integration.api.model.OffsetPaginatedListHelloWorldDto;
+import com.cerner.beadledom.jaxrs.GenericResponse;
+import io.swagger.annotations.Api;
```
johnlcox: For the client resource I think the swagger annotations can all be removed.

integration/api/src/main/java/com/cerner/beadledom/integration/api/model/HelloWorldDto.java
johnlcox: Why is the message using snake case?

integration/pom.xml
nab0310: Nope, I removed it: https://github.com/cerner/beadledom/pull/149/commits/5f16000751c242d73a0ec08af64c59f7b354e354

integration/service/src/main/java/com/cerner/beadledom/integration/service/ResteasyBootstrapModule.java
nab0310: Changed to use the List tag: https://github.com/cerner/beadledom/pull/149/commits/802c74c005968d527aef7a82390173d07569d0d5

integration/test/pom.xml
nab0310: Re-ordered: https://github.com/cerner/beadledom/pull/149/commits/0ca69d4b1c84c077044606638b2b8f9a550cf38b

integration/api/src/main/java/com/cerner/beadledom/integration/api/model/OffsetPaginatedListHelloWorldDto.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @author Nick Behrens
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@AutoValue
+@JsonDeserialize(builder = OffsetPaginatedListHelloWorldDto.Builder.class)
+public abstract class OffsetPaginatedListHelloWorldDto implements OffsetPaginatedListDto<HelloWorldDto> {
+
+  @JsonProperty("items")
+  public abstract List<HelloWorldDto> items();
```
johnlcox: Is this needed? Does it not come from the interface automatically?

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/provider/ForwardedHeaderFilterSpec.scala
nab0310: added spaces: https://github.com/cerner/beadledom/pull/149/commits/320a67c5ef71fdcbd15ae8eb0274b250433d5bee

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/provider/ForwardedHeaderFilterSpec.scala
nab0310: Yep: https://github.com/cerner/beadledom/pull/149/commits/2b4745132d5a93bb5340b45d98954dd095262209

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/provider/ForwardedHeaderFilterSpec.scala
nab0310: removed: https://github.com/cerner/beadledom/pull/149/commits/ccaf18887d21e0d9f5f736d0dd152fb8e436ecdf

integration/api/src/main/java/com/cerner/beadledom/integration/api/model/HelloWorldDto.java
nab0310: Changed it to be shorter: https://github.com/cerner/beadledom/pull/149/commits/070df335fae218738ee9473d9c6aa2abcc2c0114

integration/pom.xml
```diff
@@ -0,0 +1,103 @@
...
+
+  <modules>
+    <module>api</module>
+    <module>service</module>
+    <module>client</module>
+    <module>test</module>
+  </modules>
+
+  <properties>
+    <jetty.version>9.4.18.v20190429</jetty.version>
```
johnlcox: Looks like there is a newer version: 9.4.19.v20190610

integration/api/src/main/java/com/cerner/beadledom/integration/api/model/HelloWorldDto.java
nab0310: Yep, I changed it to break at the parameters: https://github.com/cerner/beadledom/pull/149/commits/070df335fae218738ee9473d9c6aa2abcc2c0114

integration/service/pom.xml
```diff
@@ -0,0 +1,209 @@
...
+      <groupId>javax.ws.rs</groupId>
+      <artifactId>javax.ws.rs-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-log4j12</artifactId>
```
johnlcox: Can the service use logback instead of log4j?

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/provider/ForwardedHeaderFilterSpec.scala
nab0310: Yep, I imported it and some other `Mockito` methods I used: https://github.com/cerner/beadledom/pull/149/commits/2b4745132d5a93bb5340b45d98954dd095262209

integration/test/src/docker/nginx.conf
```diff
@@ -0,0 +1,22 @@
+worker_processes 1;
+
+events { worker_connections 1024; }
+
+http {
+
+  sendfile on;
```
johnlcox: What does this do?

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/JaxRsModule.java
```diff
@@ -19,5 +26,6 @@
   protected void configure() {
     bind(CorrelationIdFilter.class).toProvider(CorrelationIdFilterProvider.class)
         .in(Singleton.class);
+    bind(ForwardedHeaderFilter.class).in(Singleton.class);
```
johnlcox: Do we have any concerns around installing this by default?

swagger2/src/main/java/com/cerner/beadledom/swagger2/SwaggerUiResource.java
```diff
@@ -44,31 +30,4 @@ public StreamingOutput getSwaggerUi(@Context UriInfo uriInfo) {
             uriInfo.getBaseUriBuilder().path(SwaggerApiResource.class).build().toASCIIString()))
     );
   }
-
```
johnlcox: This is great.

swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerUiResource.java
```diff
@@ -7,13 +7,16 @@
...
+import java.util.regex.Pattern;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.Context;
+import javax.ws.rs.core.HttpHeaders;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
+import javax.ws.rs.core.SecurityContext;
```
johnlcox: Are these imports unused? I don't see any other changes in this file.

integration/service/pom.xml
```diff
@@ -0,0 +1,209 @@
...
+              <alias>beadledom-integration-service</alias>
+              <name>${project.groupId}/${project.artifactId}:${project.version}</name>
+
+              <build>
+                <cleanup>remove</cleanup>
+                <env>
+                  <PROJECT_VERSION>${project.version}</PROJECT_VERSION>
+                </env>
+                <ports>
+                  <port>8080</port>
```
johnlcox: Can the port be randomized? Is there any disadvantage to keeping it just 8080?

integration/api/src/main/java/com/cerner/beadledom/integration/api/HelloWorldResource.java
```diff
@@ -0,0 +1,35 @@
...
+public interface HelloWorldResource {
+
+  /**
+   * A endpoint that returns a paginated list of {@link HelloWorldDto}.
+   *
+   * @return an {@link OffsetPaginatedListHelloWorldDto}
+   */
+  @ApiOperation(
+      value = "Retrieves hello world data.",
+      response = HelloWorldDto.class)
```
charanpanchagnula: Shouldn't response here be OffsetPaginatedListHelloWorldDto ?

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/ForwardedHeaderFilter.java
```diff
@@ -0,0 +1,72 @@
...
+        && matcher.group("protocolValue").contains("https");
+  }
+
+  /**
+   * Checks the X-Forwarded-Proto header for a value of https.
+   * @param requestContext The {@link ContainerRequestContext} object containing the container request headers.
+   * @return true if the `X-Forwarded-Proto` header is present and contains https; false otherwise.
+   */
+  private boolean hasSecureXForwardedProtoHeader(ContainerRequestContext requestContext) {
+    return "https".equals(requestContext.getHeaderString("X-Forwarded-Proto"));
```
nathanschile: If the header is `X-Forwarded-Proto` is lower case (`x-forwarded-proto`) does this still work?

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/ForwardedHeaderFilter.java
```diff
@@ -0,0 +1,72 @@
...
+        || hasSecureXForwardedProtoHeader(requestContext);
+  }
+
+  /**
+   * Checks the Forwarded header for a protocol value of https.
+   * @param requestContext The {@link ContainerRequestContext} object containing the container request headers.
+   * @return true if the `Forwarded` header is present and contains https; false otherwise.
+   */
+  private boolean hasSecureForwardedHeader(ContainerRequestContext requestContext) {
+    String forwardedHeader = requestContext.getHeaderString("Forwarded");
```
nathanschile: Do we need to support the "Forwarded" header? It isn't the standard, https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded

integration/api/src/main/java/com/cerner/beadledom/integration/api/HelloWorldResource.java
```diff
@@ -0,0 +1,35 @@
...
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+/**
+ * Client interface of an offset based paginated list endpoint resource.
+ *
+ * @author Nick Behrens
+ */
+@Api(value = "/hello", description = "Retrieve hello world data")
```
charanpanchagnula: description is deprecated

integration/service/src/main/java/com/cerner/beadledom/integration/service/resource/HelloWorldResource.java
```diff
@@ -0,0 +1,43 @@
...
+
+  /**
+   * An endpoint that returns a paginated list of {@link HelloWorldDto}.
+   *
+   * @return an {@link OffsetPaginatedList} containing {@link HelloWorldDto}'s.
+   */
+  @ApiOperation(
+      value = "Retrieves hello world data.",
+      response = HelloWorldDto.class)
```
charanpanchagnula: Shouldn't response here be OffsetPaginatedListHelloWorldDto or OffsetPaginatedList<HelloWorldDto>?

integration/service/src/main/java/com/cerner/beadledom/integration/service/resource/HelloWorldResource.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+/**
+ * Server implementation of an offset based paginated list endpoint.
+ *
+ * @author Nick Behrens
+ */
+@Api(value = "/hello", description = "Retrieve hello world data")
```
charanpanchagnula: description is deprecated

integration/service/src/main/java/com/cerner/beadledom/integration/service/resource/HelloWorldResource.java
charanpanchagnula: This is how it works in red-tape also for v2 resources. Whats the reason for this discrepancy?

integration/api/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+
+  <parent>
+    <artifactId>beadledom-integration</artifactId>
+    <groupId>com.cerner.beadledom</groupId>
+    <version>3.3-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>beadledom-integration-api</artifactId>
+
+  <dependencies>
```
charanpanchagnula: I think there are some used but undeclared dependencies found (you'll be able to find out when you run `mvn dependency:analyze -pl api/`

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/ForwardedHeaderFilter.java
nab0310: So the weird thing with that documentation is that if you go to the link for the `X-Forwarded-x` headers, it says that "A standardized version of this header is the HTTP `Forwarded` header." So I decided to just support both, the documentation was confusing.

integration/service/src/main/java/com/cerner/beadledom/integration/service/ResteasyBootstrapModule.java
```diff
@@ -0,0 +1,59 @@
...
+    BuildInfo buildInfo = BuildInfo.load(ResteasyBootstrapModule.class
+        .getResourceAsStream("build-info.properties"));
+    bind(BuildInfo.class).toInstance(buildInfo);
+    bind(ServiceMetadata.class).toInstance(ServiceMetadata.create(buildInfo));
+
+    OffsetPaginationConfiguration paginationConfig = OffsetPaginationConfiguration.builder()
+        .setDefaultLimit(10)
+        .setMaxLimit(10)
+        .build();
+    OptionalBinder
```
charanpanchagnula: Possibly stupid question, why do we need to use OptionalBinder here?

integration/service/src/main/resources/com/cerner/beadledom/integration/service/build-info.properties
```diff
@@ -0,0 +1,4 @@
+git.commit.id=${git.commit.id}
```
charanpanchagnula: This is for my knowledge, why do we need this file?

integration/test/src/docker/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM nginx
+
```
charanpanchagnula: Could we add some documentation in this Dockerfile to understand whats going on here?

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/ForwardedHeaderFilter.java
```diff
@@ -0,0 +1,72 @@
...
+    requestContext.setRequestUri(
+        requestContext.getUriInfo().getRequestUriBuilder().scheme(httpScheme).build());
+  }
+
+  /**
+   * Checks if the request was made from a secure context or not.
+   * @param requestContext The {@link ContainerRequestContext} object containing the container request headers.
+   * @return true if the request was made from a secure context, false otherwise.
+   */
+  private boolean requestIsSecure(ContainerRequestContext requestContext) {
```
charanpanchagnula: Could we change the method name to isRequestSecure

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/ForwardedHeaderFilter.java
```diff
@@ -0,0 +1,72 @@
...
+import javax.ws.rs.container.PreMatching;
+import javax.ws.rs.ext.Provider;
+
+/**
+ * The ForwardedHeaderFilter reads the request context and determines if the client request was made
+ * from a secure protocol (HTTPS). If the request originated from secure context, it updates the
+ * request to reflect the original protocol.
+ *
+ * <p>To determine if the client request made in a secure context was forwarded by a load balancer
+ * or proxy, the "Forwarded" and "X-Forwarded-Proto" values are extracted and checked.</p>
```
charanpanchagnula: It seems like it is deemed a secure context if one of the headers is present in the request and not both of them. Should we explicitly say that here in the description?

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/ForwardedHeaderFilter.java
```diff
@@ -0,0 +1,72 @@
...
+   * Checks the Forwarded header for a protocol value of https.
+   * @param requestContext The {@link ContainerRequestContext} object containing the container request headers.
+   * @return true if the `Forwarded` header is present and contains https; false otherwise.
+   */
+  private boolean hasSecureForwardedHeader(ContainerRequestContext requestContext) {
+    String forwardedHeader = requestContext.getHeaderString("Forwarded");
+    if (forwardedHeader == null) {
+      return false;
+    }
+    Pattern forwardedPairs = Pattern.compile("proto=(?<protocolValue>[^;]*)(;|\\z)");
```
charanpanchagnula: could we add a doc explaining this regex because its not clear

johnlcox: Should it also be extracted out to the class level so the pattern isn't recreated on every request?

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/ForwardedHeaderFilter.java
nathanschile: Looks like `Forwarded` is the standard, https://tools.ietf.org/html/rfc7239 :)

integration/api/src/main/java/com/cerner/beadledom/integration/api/HelloWorldResource.java
nab0310: Yes, it should be alright to go there: https://github.com/cerner/beadledom/pull/149/commits/acc0165095c733471eec959e3a986d962a3146e2

integration/api/src/main/java/com/cerner/beadledom/integration/api/HelloWorldResource.java
nab0310: I removed them when I moved the resource to the client: https://github.com/cerner/beadledom/pull/149/commits/acc0165095c733471eec959e3a986d962a3146e2

integration/api/src/main/java/com/cerner/beadledom/integration/api/model/HelloWorldDto.java
nab0310: Because that's what beadledom generated when I ran the `beadledom new` command. I changed it to use camel: https://github.com/cerner/beadledom/pull/149/commits/f8f68c08bbc29dbb7ca87ef6c3e25de9dd562f23

integration/pom.xml
nab0310: There is: https://github.com/cerner/beadledom/pull/149/commits/17f663e0b1f0011fdba37d921c18f6e2e30788bc

integration/api/src/main/java/com/cerner/beadledom/integration/api/model/OffsetPaginatedListHelloWorldDto.java
nab0310: It isn't needed, good catch: https://github.com/cerner/beadledom/pull/149/commits/2c3d8ca7133d1342f729c18d5e2ef5d56ae33db5

integration/test/src/docker/nginx.conf
nab0310: This allows nginx to send chunks of a file directly over the wire rather than copying it to a buffer first. It's a minor performance optimization that we probably don't need.

integration/service/pom.xml
nab0310: I'm not sure it could be randomized, I use this port to redirect the traffic through nginx. I don't think I can use environment variables inside of the nginx.config files so it might have to be hardcoded. There could be a disadvantage in hardcoding the port because it makes the tests rather brittle (ie. if any other is running on port 8080 it would fail). I'll can look into using an environment variable.

integration/service/src/main/java/com/cerner/beadledom/integration/service/resource/HelloWorldResource.java
nab0310: It ideally should be `OffsetPaginatedList<HelloWorldDto>` but swagger annotations don't support generics and we can't use the built in support because we are actually returning a `GenericResponse<OffsetPaginatedList<HelloWorldDto>>` which is way messier of a swagger model then just a HelloWorldDto is.

integration/service/src/main/java/com/cerner/beadledom/integration/service/ResteasyBootstrapModule.java
nab0310: I'm not quite sure, it was what was in the documentation for installing pagination so that's why I did it. @johnlcox, is there any good reason to use an Optional Binder here?

integration/service/src/main/resources/com/cerner/beadledom/integration/service/build-info.properties
nab0310: This file is used by the ServiceMetadata to help swagger get some metadata about the service. It's used here: https://github.com/cerner/beadledom/pull/149/files/070df335fae218738ee9473d9c6aa2abcc2c0114#diff-fad1ad5b719cd2aebc6b4b10bdcfc825R34

nab0310: Come to think of it, we only really need the project version out of this file, so we might be able to remove all the other key value pairs....

integration/api/pom.xml
nab0310: I checked this and other submodules and removed and added some dependencies: https://github.com/cerner/beadledom/pull/149/commits/9ea1691d731343e157a568614742303310dd6651

I'll run this again before I commit as a final sanity check.

integration/service/src/main/java/com/cerner/beadledom/integration/service/resource/HelloWorldResource.java
nab0310: It is, I just removed it. There didn't seem to be a viable replacement, but this is just an example service so I'm not too worried about the swagger docs being good: https://github.com/cerner/beadledom/pull/149/commits/0cfffed54c3dbbcc7b53efa136c19292644441fb

integration/test/src/docker/nginx.conf
nab0310: Removed: https://github.com/cerner/beadledom/pull/149/commits/8a3c86cbe865def41f7369374be99da2c50fde2b

swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerUiResource.java
nab0310: They are.. I'm not sure why I didn't get a checkstyle issue when I ran `mvn clean install`. Removed here: https://github.com/cerner/beadledom/pull/149/commits/61ab77af9cf42368041d0268f9d488dc890d3d48

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/ForwardedHeaderFilter.java
nab0310: Yep, it should be. I moved it out and added a little documentation on what it does here: https://github.com/cerner/beadledom/pull/149/commits/9fd05e06fd0a408492e73aabea311542f32f36ac

swagger1/src/main/java/com/cerner/beadledom/swagger/SwaggerUiResource.java
nab0310: They are.. I'm not sure why I didn't get a checkstyle issue when I ran mvn clean install. Removed here: 61ab77a

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/ForwardedHeaderFilter.java
nab0310: Even though `Forwarded` is the standard, I'm not sure what some other proxy setups would use. I'm a little more comfortable with leaving the support in there and making a note of the deprecation in the documentation.

integration/service/src/main/java/com/cerner/beadledom/integration/service/ResteasyBootstrapModule.java
charanpanchagnula: Could you link me to this documentation?

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/ForwardedHeaderFilter.java
nab0310: It does, I verified it by setting the integration reverse proxy to use a lower case (x-forwarded-proto) header. I committed the changed nginx config: https://github.com/cerner/beadledom/pull/149/commits/f4695a52f8ff844164ee7e8b59263a32995bae32

integration/service/src/main/java/com/cerner/beadledom/integration/service/ResteasyBootstrapModule.java
nab0310: https://github.com/cerner/beadledom/blob/master/docs/source/manual/pagination.rst

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/ForwardedHeaderFilter.java
nab0310: I added a little clearer description: https://github.com/cerner/beadledom/pull/149/commits/8930d43b5dab6c50fc6b30305109bea73744d62a

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/ForwardedHeaderFilter.java
nab0310: Yep, that makes more sense to me: https://github.com/cerner/beadledom/pull/149/commits/6cf5d63ab23dae4bcb9d740a353363aa64900e01

integration/test/src/docker/Dockerfile
nab0310: https://github.com/cerner/beadledom/pull/149/commits/00f709820b4beb51b556119aceb9ecd6bfdbe7c6

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/JaxRsModule.java
nab0310: Yeah, we should. It would be safer to just let the clients install it if they want to.

nab0310: I removed the default install, and added docs in the jaxrs.rst file: https://github.com/cerner/beadledom/pull/149/commits/5bce5a5cc4b901cf9b67fa7953618919bb44b789

integration/client/src/main/java/com/cerner/beadledom/integration/client/IntegrationClientFeature.java
nab0310: yup: https://github.com/cerner/beadledom/pull/149/commits/934f3528830366a99897f34926cc5ff9398dd98a

integration/service/pom.xml
nab0310: I changed the service to use logback, but I couldn't get request logging to work: https://github.com/cerner/beadledom/pull/149/commits/3e525a5830d72c608607bef3c1b2144559b2405d

integration/service/src/main/java/com/cerner/beadledom/integration/service/resource/HelloWorldResource.java
charanpanchagnula: Then shouldn't it just be OffsetPaginatedList ?

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/ForwardedHeaderFilter.java
```diff
@@ -0,0 +1,79 @@
...
+@Provider
+@Priority(Priorities.HEADER_DECORATOR)
+@PreMatching
+public class ForwardedHeaderFilter implements ContainerRequestFilter {
+  // This regex takes the value of the 'proto=' field and sets it to a group. The 'proto=' field can
+  // be followed by either another key=value pair separated by a ; or it could be the end of the header
+  // Examples:
+  // proto=https => 'protocolValue' group is set to "https"
+  // proto=http;host=localhost => 'protocolValue' group is set to "http"
+  Pattern forwardedPairs = Pattern.compile("proto=(?<protocolValue>[^;]*)(;|\\z)");
```
charanpanchagnula: can this be private?

nab0310: yep: https://github.com/cerner/beadledom/pull/149/commits/e28e1da6f2cae0e5ba16371329a4caa0e663684e

integration/service/src/main/java/com/cerner/beadledom/integration/service/resource/HelloWorldResource.java
nab0310: So with the way the swagger annotations work, generics aren't represented at all. So putting `OffsetPaginatedList<HelloWorldDto>` in there is out of the question. I would prefer it to be a `HelloWorldDto` because thats the actual object as opposed to just the list interface that is wrapping it.

integration/service/src/main/java/com/cerner/beadledom/integration/service/resource/HelloWorldResource.java
nab0310: It's because the `OffsetPaginatedList` contains logic to generate the next and previous links. The Dto however doesn't contain this logic because it doesn't need to. We return the `OffsetPaginatedList` because the links for the returned object matter but for the dto representation they don't.

nab0310: I'm not entirely sure, the docs for `OffsetPagiantedListDto` say that "The `OffsetPaginatedListDto` is used on the client side instead of {@link OffsetPaginatedList} to circumvent type erasure."

PR #155, opened 2019-09-26
Upgrade jackson to 2.10

nab0310:

Jackson 2.9.9 had a security vulnerability.

### How was it tested?

mvn clean install

### How to test


johnlcox:
Looks like `commons-beanutils` needs to be upgraded to 1.9.4 as well.

PR #158, opened 2019-10-07
Update ThreadLocalCommitHookExecutingTransactionListener to override TransactionListener#beginStart instead of TransactionListener#beginEnd

nathanschile:

If there is a exception while beginning the [transaction](https://github.com/jOOQ/jOOQ/blob/version-3.11.12/jOOQ/src/main/java/org/jooq/impl/DefaultDSLContext.java#L516) then TransactionListener#beginEnd
is never called. When this happened, ThreadLocalCommitHookExecutingTransactionListener#rollbackEnd would throw a NoSuchElementException since TransactionListener#beginEnd was never called setting up transactionDeques object

### How was it tested?

Updated existing tests.

### How to test

### DIFF
--- a/jooq/src/main/java/com/cerner/beadledom/jooq/ThreadLocalCommitHookExecutingTransactionListener.java
+++ b/jooq/src/main/java/com/cerner/beadledom/jooq/ThreadLocalCommitHookExecutingTransactionListener.java
@@ -25,7 +25,7 @@ class ThreadLocalCommitHookExecutingTransactionListener extends DefaultTransacti
   }
 
   @Override
-  public void beginEnd(TransactionContext ctx) {
+  public void beginStart(TransactionContext ctx) {
     JooqTransaction transaction = new JooqTransaction();
 
     Deque<JooqTransaction> transactions = transactions();
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Beadledom Changelog
 
+## [Unreleased]
+
+### Fixed
+* Update `ThreadLocalCommitHookExecutingTransactionListener` to override `TransactionListener#beginStart` instead of `TransactionListener#beginEnd`
+
 ## 3.3 - 8 August 2019
 
 ### Added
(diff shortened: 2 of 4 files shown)


jooq/src/main/java/com/cerner/beadledom/jooq/ThreadLocalCommitHookExecutingTransactionListener.java
```diff
@@ -25,7 +25,7 @@
   }
```
nathanschile: This is the exception that exposed the issue

```
Caused by: java.util.NoSuchElementException: null
	at java.util.ArrayDeque.removeFirst(ArrayDeque.java:285)
	at java.util.ArrayDeque.pop(ArrayDeque.java:522)
	at com.cerner.beadledom.jooq.ThreadLocalCommitHookExecutingTransactionListener.rollbackEnd(ThreadLocalCommitHookExecutingTransactionListener.java:64)
	at org.jooq.impl.TransactionListeners.rollbackEnd(TransactionListeners.java:93)
	at org.jooq.impl.DefaultDSLContext.lambda$transactionResult0$0(DefaultDSLContext.java:555)
	at org.jooq.impl.DefaultDSLContext.transactionResult0(DefaultDSLContext.java:574)
	at org.jooq.impl.DefaultDSLContext.transactionResult(DefaultDSLContext.java:481)
	at com.cerner.beadledom.jooq.JooqTxnInterceptor.invokeInTransactionAndUnitOfWork(JooqTxnInterceptor.java:47)
	at com.cerner.beadledom.jooq.JooqTxnInterceptor.invoke(JooqTxnInterceptor.java:27)
```

venki09: Do we need to update changelog?

nathanschile: Yup, added https://github.com/cerner/beadledom/pull/158/commits/427bc17abf71852fffbfd5c6f56da70684509133

PR #173, opened 2020-02-13
Add basic retry backoff for HTTP client request exceptions

johnlcox:

Added an exponential retry delay between HTTP request retries for IOExceptions.

### How was it tested?

New unit tests.

### How to test

### DIFF
--- a/client/resteasy-client/src/main/java/com/cerner/beadledom/client/resteasy/ExponentialBackoffHttpRequestRetryHandler.java
+++ b/client/resteasy-client/src/main/java/com/cerner/beadledom/client/resteasy/ExponentialBackoffHttpRequestRetryHandler.java
@@ -25,7 +25,7 @@ class ExponentialBackoffHttpRequestRetryHandler extends StandardHttpRequestRetry
     if (!super.retryRequest(exception, executionCount, context)) {
       return false;
     }
-    
+
     // Adding the delay here is not ideal, but there doesn't appear to be a better place in Apache
     // httpclient similar to ServiceUnavailableRetryStrategy#getRetryInterval.
     try {
--- a/client/resteasy-client/src/main/java/com/cerner/beadledom/client/resteasy/ExponentialBackoffHttpRequestRetryHandler.java
+++ b/client/resteasy-client/src/main/java/com/cerner/beadledom/client/resteasy/ExponentialBackoffHttpRequestRetryHandler.java
@@ -1,0 +1,45 @@
+package com.cerner.beadledom.client.resteasy;
+
+import java.io.IOException;
+import java.util.Random;
+import org.apache.http.impl.client.StandardHttpRequestRetryHandler;
+import org.apache.http.protocol.HttpContext;
+
+/**
+ * A {@link StandardHttpRequestRetryHandler} that adds an exponential back-off to retries.
+ *
+ * @author John Leacox
+ * @since 3.5
+ */
+class ExponentialBackoffHttpRequestRetryHandler extends StandardHttpRequestRetryHandler {
+  private static final int INITIAL_DELAY = 100;
+  private static final int BACKOFF_FACTOR = 2;
+  private static final Random random = new Random();
+
+  ExponentialBackoffHttpRequestRetryHandler(int retryCount, boolean requestSentRetryEnabled) {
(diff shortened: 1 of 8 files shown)


client/resteasy-client/src/main/java/com/cerner/beadledom/client/resteasy/ExponentialBackoffHttpRequestRetryHandler.java
```diff
@@ -0,0 +1,45 @@
...
+  ExponentialBackoffHttpRequestRetryHandler(int retryCount, boolean requestSentRetryEnabled) {
+    super(retryCount, requestSentRetryEnabled);
+  }
+
+  @Override
+  public boolean retryRequest(IOException exception, int executionCount, HttpContext context) {
+    // Adding the delay here is not ideal, but there doesn't appear to be a better place in Apache
+    // httpclient similar to ServiceUnavailableRetryStrategy#getRetryInterval.
+    try {
+      Thread.sleep(calculateRetryDelay(executionCount));
```
nathanschile: Don't you want to call `super.retryRequest(exception, executionCount, context);` before sleeping, and only sleep if the request should be retried?

johnlcox: Yes, good point. https://github.com/cerner/beadledom/pull/173/commits/e421982ad45b6c0f606cbd8b4f76bbc8c14d5827#diff-2966b7f567e51e0636255711fa0596e7R25

client/resteasy-client/src/main/java/com/cerner/beadledom/client/resteasy/ExponentialBackoffHttpRequestRetryHandler.java
```diff
@@ -0,0 +1,49 @@
...
+  private static final int INITIAL_DELAY = 100;
+  private static final int BACKOFF_FACTOR = 2;
+  private static final Random random = new Random();
+
+  ExponentialBackoffHttpRequestRetryHandler(int retryCount, boolean requestSentRetryEnabled) {
+    super(retryCount, requestSentRetryEnabled);
+  }
+
+  @Override
+  public boolean retryRequest(IOException exception, int executionCount, HttpContext context) {
```
nab0310: Should we respect a `Retry-After` header if there is one on the request? 
https://tools.ietf.org/html/rfc7231#section-7.1.3

client/resteasy-client/src/main/java/com/cerner/beadledom/client/resteasy/ExponentialBackoffHttpRequestRetryHandler.java
```diff
@@ -0,0 +1,49 @@
...
+      Thread.sleep(calculateRetryDelay(executionCount));
+    } catch (InterruptedException e) {
+      // ignore
+    }
+
+    return true;
+  }
+
+  // VisibleForTesting
+  int calculateRetryDelay(int executionCount) {
```
nab0310: Should there be a maximum delay between requests?

client/resteasy-client/src/main/java/com/cerner/beadledom/client/resteasy/ExponentialBackoffHttpRequestRetryHandler.java
johnlcox: I think handling of that would be in the Service Unavailable retry strategy. This retry handler is for IOExceptions, so there won't be any response or headers to look at.

client/resteasy-client/src/main/java/com/cerner/beadledom/client/resteasy/ExponentialBackoffHttpRequestRetryHandler.java
johnlcox: If we were implementing the full internal RFD on HTTP retries, then we would want a maximum, but my intent with this PR is to add a very simple retry backoff that follows the RFD but doesn't fully implement it. 

This class will only retry 3 times and is not configurable, so the maximum delay from the internal RFD can never be reached.

PR #175, opened 2020-03-12
Deprecate beadledom-client

### LINKED ISSUE #160: Deprecate beadledom-client
Deprecate beadledom-client in favor of Square's [Retrofit](https://github.com/square/retrofit/)

CalebBassham:

Marked `beadledom-client` module as deprecated in favor of Retrofit. Added notes indicating such to the docs of beadledom-client, the 4.0 migration guide, and the readme.

Closes #160

### DIFF
--- a/client/README.rst
+++ b/client/README.rst
@@ -1,4 +1,6 @@
 .. _beadledom-client:
 
+Deprecated since 4.0, use `Retrofit <https://github.com/square/retrofit>`_ instead.
+
 `Check out our docs for more information <http://cerner.github.io/beadledom>`_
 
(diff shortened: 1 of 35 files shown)


client/README.rst
```diff
@@ -1,4 +1,6 @@
 .. _beadledom-client:
 
+Deprecated since 4.0, use `Retrofit <https://github.com/square/retrofit>`_ instead.
```
johnlcox: Do we want to deprecate this in a 3.x release even?

CalebBassham: Would be then be removing it from the 4.0 release?

johnlcox: I wish we could do that, but we probably have too many consumers to remove it in 4.0, so it would just be deprecated in both. Maybe re-evaluate removal in 5.0. Thoughts, @nathanschile?

nathanschile: I agree with @johnlcox We want people to upgrade to 4.0, so lets leave it in.

johnlcox:
I squashed the changes from this PR and open a new PR for Beadledom 3.x: https://github.com/cerner/beadledom/pull/177

PR #176, opened 2020-04-01
Upgrade to Scala 2.12 and scalatest 3.1

johnlcox:

Upgrades to Scala 2.12 and scalatest 3.1 so that we can be on the latest scalatest version. We still can't go to Scala 2.13 because some of our internally used Scala dependencies (for example gatling) only support 2.12 so far.

### How was it tested?

Existing tests

### How to test

### DIFF
--- a/docs/source/releases/Beadledom40.rst
+++ b/docs/source/releases/Beadledom40.rst
@@ -64,11 +64,57 @@ Scalatest provides autofix mechanisms for this upgrade, but requires multiples s
           </dependency>
         </dependencies>
       </plugin>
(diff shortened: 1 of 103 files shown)


docs/source/releases/Beadledom40.rst
```diff
@@ -28,3 +28,93 @@ on beadledom-swagger1 to beadledom-swagger2. The user guide for beadledom-swagge
 
 The beadledom-stagemonitor module was removed. This will require updating your poms that have a dependency
 on beadledom-stagemonitor.
+
```
nab0310, line 31: Good thinking to document the changes and fixes you had to do for consumers.

PR #177, opened 2020-04-02
Deprecate beadledom-client

### LINKED ISSUE #160: Deprecate beadledom-client

johnlcox:

Marked beadledom-client modules and classes as deprecated. We are now recommending the use of Retrofit for HTTP clients.

Closes #160


nathanschile:
> Should we update the archetype to remove the client module as part of this change too?

We probably should.

johnlcox:
Archetype updated here: https://github.com/cerner/beadledom/pull/177/commits/60cc047b2db1ea2589fd6797c1ff570f52122c8d

PR #179, opened 2020-04-13
Resteasy 4 uplift

venki09:
* Uplift to resteasy-4

### How was it tested?
* Tested with a service locally
* Unit and integration tests.

### How to test

### DIFF
--- a/client/beadledom-client-test/pom.xml
+++ b/client/beadledom-client-test/pom.xml
@@ -39,6 +39,11 @@
 
     <!--Test Dependencies-->
     <dependency>
+      <groupId>javax.servlet</groupId>
+      <artifactId>javax.servlet-api</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>org.apache.tomcat</groupId>
       <artifactId>tomcat-jasper</artifactId>
       <scope>test</scope>
--- a/client/pom.xml
+++ b/client/pom.xml
@@ -43,21 +43,6 @@
         <artifactId>beadledom-client-example-service</artifactId>
         <version>${project.version}</version>
       </dependency>
-      <dependency>
-        <groupId>org.jboss.resteasy</groupId>
-        <artifactId>resteasy-client</artifactId>
-        <version>${resteasy.version}</version>
-      </dependency>
-      <dependency>
-        <groupId>org.jboss.resteasy</groupId>
-        <artifactId>resteasy-guice</artifactId>
-        <version>${resteasy.version}</version>
-      </dependency>
-      <dependency>
-        <groupId>org.jboss.resteasy</groupId>
-        <artifactId>resteasy-jaxrs</artifactId>
-        <version>${resteasy.version}</version>
-      </dependency>
     </dependencies>
   </dependencyManagement>
 </project>
--- a/client/resteasy-client/src/test/scala/com/cerner/beadledom/client/resteasy/ResteasyClientSpec.scala
+++ b/client/resteasy-client/src/test/scala/com/cerner/beadledom/client/resteasy/ResteasyClientSpec.scala
@@ -28,17 +28,17 @@ class ResteasyClientSpec(contextRoot: String, servicePort: Int)
           classOf[BeadledomResteasyWebTarget]
     }
 
-//    it("re-uses the service request correlationId when it is present") {
-//      val captureFilter = new CaptureCorrelationIdFilter()
-//
-//      val client = BeadledomResteasyClientBuilder.newBuilder()
-//          .register(captureFilter, Array[Class[_]](classOf[ClientRequestFilter]): _*)
-//          .build()
-//      val proxy = client.target(s"http://localhost:$servicePort/$contextRoot")
-//          .proxy(classOf[TestResource])
-//
-//      proxy.loopyGetCorrelationId() must be(captureFilter.getCapturedCorrelationId)
-//    }
+    it("re-uses the service request correlationId when it is present") {
+      val captureFilter = new CaptureCorrelationIdFilter()
+
+      val client = BeadledomResteasyClientBuilder.newBuilder()
+          .register(captureFilter, Array[Class[_]](classOf[ClientRequestFilter]): _*)
+          .build()
+      val proxy = client.target(s"http://localhost:$servicePort/$contextRoot")
+          .proxy(classOf[TestResource])
+
+      proxy.loopyGetCorrelationId() must be(captureFilter.getCapturedCorrelationId)
+    }
 
     it("falls back to the MDC correlationId when the header is not present") {
       MDC.put(CorrelationIdContext.DEFAULT_HEADER_NAME,  "mdcFallbackId")
--- a/client/resteasy-client/src/test/scala/com/cerner/beadledom/client/resteasy/ResteasyClientSpec.scala
+++ b/client/resteasy-client/src/test/scala/com/cerner/beadledom/client/resteasy/ResteasyClientSpec.scala
@@ -1,12 +1,10 @@
 package com.cerner.beadledom.client.resteasy
 
-import javax.ws.rs.client.ClientRequestFilter
 import com.cerner.beadledom.client._
-
 import com.fasterxml.jackson.annotation.JsonInclude
 import com.fasterxml.jackson.databind.{DeserializationFeature, ObjectMapper, PropertyNamingStrategy, SerializationFeature}
 import com.fasterxml.jackson.jaxrs.json.JacksonJsonProvider
-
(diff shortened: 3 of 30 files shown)


client/resteasy-client/src/test/scala/com/cerner/beadledom/client/resteasy/ResteasyClientSpec.scala
```diff
@@ -30,20 +28,20 @@ class ResteasyClientSpec(contextRoot: String, servicePort: Int)
...
-
-      val client = BeadledomResteasyClientBuilder.newBuilder()
-          .register(captureFilter, Array[Class[_]](classOf[ClientRequestFilter]): _*)
-          .build()
-      val proxy = client.target(s"http://localhost:$servicePort/$contextRoot")
-          .proxy(classOf[TestResource])
-
-      proxy.loopyGetCorrelationId() must be(captureFilter.getCapturedCorrelationId)
-    }
+//    it("re-uses the service request correlationId when it is present") {
```
venki09, line 31: I ll rebase my branch with changes from https://github.com/cerner/beadledom/pull/180 to get out of this issue

client/pom.xml
```diff
@@ -48,14 +48,24 @@
         <artifactId>resteasy-client</artifactId>
         <version>${resteasy.version}</version>
       </dependency>
+      <dependency>
+        <groupId>org.jboss.resteasy</groupId>
+        <artifactId>resteasy-client-api</artifactId>
+        <version>${resteasy.version}</version>
```
nathanschile, line 54: The version is already managed in the parent pom, we shouldn't also need to do it here.

pom.xml
```diff
@@ -448,12 +448,22 @@
       </dependency>
       <dependency>
         <groupId>org.jboss.resteasy</groupId>
-        <artifactId>resteasy-guice</artifactId>
+        <artifactId>resteasy-client-api</artifactId>
+        <version>${resteasy.version}</version>
```
nathanschile, line 452: Lets use the resteasy-bom

```
<dependency>
    <groupId>org.jboss.resteasy</groupId>
    <artifactId>resteasy-bom</artifactId>
    <version>4.5.3.Final</version>
    <type>pom</type>
</dependency>
```

client/beadledom-client-test/pom.xml
```diff
@@ -38,6 +38,11 @@
     </dependency>
```
nathanschile, line 39: Are you intending this to go in both 3.x and 4.x of Beadledom?

resteasy/src/main/java/com/cerner/beadledom/resteasy/GzipContentEncodingFilter.java
```diff
@@ -15,7 +15,7 @@
 @Provider
 @Priority(Priorities.HEADER_DECORATOR)
 public class GzipContentEncodingFilter extends ServerContentEncodingAnnotationFilter {
-  public GzipContentEncodingFilter() {
+  GzipContentEncodingFilter() {
```
nathanschile, line 18: Why did you narrow the scope?

venki09: Oops. I was just testing something. Reverted here https://github.com/cerner/beadledom/pull/179/commits/fb70f680b7583aaf63887cbeb4987261b9fcdc4b

pom.xml
venki09: 224d842a I still had to pull in `resteasy-core` separately as that was not part of the bom. As per the migration guide, https://docs.jboss.org/resteasy/docs/4.0.0.Final/userguide/html/Migration_from_older_versions.html it seems like we shouldn
t directly rely on classes from the resteasy-core module. But we rely on that for tests I think. Is that fine?

client/pom.xml
venki09: This is dependency management section. I am getting errors if I don't specify version here. Am I missing something trivial here? 🤔

nathanschile: You need to add 

      <dependency>
        <groupId>org.jboss.resteasy</groupId>
        <artifactId>resteasy-client-api</artifactId>
        <version>${resteasy.version}</version>
      </dependency>

to the parent pom, then you can remove the resteasy dependencies from here.

pom.xml
nathanschile: Looking at the code, I don't see a way for us not to use resteasy-core. I would log a issue/pr to resteasy to have the dependencies that are not managed by the bom, to be.

client/resteasy-client/src/test/scala/com/cerner/beadledom/client/resteasy/ResteasyClientSpec.scala
johnlcox: Why doesn't this test work after the upgrade? Wouldn't we want to test both the fallback to `ResteasyContext` and the new correlation ID change in #180?

venki09: The reason this test is not working is because during each client call, the context is getting cleared, which means, previosuly set correlationID is cleared. This is a problem if multiple clients are created and called inside a single JAX-RS transaction.

resteasy/src/main/java/com/cerner/beadledom/resteasy/ResteasyModule.java
```diff
@@ -7,7 +7,7 @@
 import javax.inject.Singleton;
 import javax.servlet.ServletContext;
 import org.jboss.resteasy.plugins.guice.ext.RequestScopeModule;
-import org.jboss.resteasy.plugins.interceptors.encoding.GZIPEncodingInterceptor;
+import org.jboss.resteasy.plugins.interceptors.GZIPEncodingInterceptor;
```
johnlcox, line 10: I'm having a funny memory (not sure if true or not) that this class was originally here and then in one of the 3.x releases it got moved into `encoding` and now they are moving it back.

client/beadledom-client-test/pom.xml
venki09: No this is only intended for 4.x release

dependency-check-suppressions.xml
```diff
@@ -10,13 +10,22 @@
     <cve>CVE-2012-6708</cve>
     <cve>CVE-2015-9251</cve>
     <cve>CVE-2019-11358</cve>
+    <vulnerabilityName>Regex in its jQuery.htmlPrefilter  sometimes may introduce XSS</vulnerabilityName>
```
johnlcox, line 13: All of the changes in this file won't be necessary when compared against the 4.0 branch since it removes this dependencies.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmapping/WebApplicationExceptionMapper.java
```diff
@@ -46,14 +46,16 @@ public Response toResponse(WebApplicationException exception) {
       logger.error("An unhandled exception was thrown.", exception);
     }
 
+    JsonError errorBody = response != null && response.getEntity() instanceof JsonError ? response
```
johnlcox, line 49: Is this a fix for an existing bug or did something with the upgrade lead to this being necessary?

dependency-check-suppressions.xml
venki09: yeah I will have to open a PR once #180 is merged. While doing that, I will do it against the 4.x branch

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmapping/WebApplicationExceptionMapper.java
venki09: Yeah, from resteasy 3.10+, the ExceptionHandler class of resteasy was updated to look for available exception mappers for a class and its super class. In one of my local services, I am throwing a BadRequestException with a JsonError entity which is a subClass of the WebApplicationException so this mapper catches it. So I have updated this mapper to honor any existing JsonError entity that could already be set by the consumer.

client/pom.xml
venki09: 999b146f

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/exceptionmapping/WebApplicationExceptionMapper.java
venki09: 7701ab8f I have refactored this to read a little better

venki09:
I am going to close this PR and open a new one against the 4.x beadledom branch.

Issue #166, opened 2020-01-14
Extract CorrelationIdContext out of the client modules


johnlcox:
The `CorrelationIdContext` class should have minimal dependencies and be usable both from server-side code and client-side code. The server-side code should include a filter to set the incoming correlation id header into the context and unset it after the request.

PR #180, opened 2020-04-13
Create Correleation Module

nathanschile:
Created Correlation Module. https://github.com/cerner/beadledom/issues/166

### How was it tested?

Tested using unit tests.

### How to test


correlation/src/test/java/com/cerner/beadledom/correlation/CorrelationIdContextTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import com.cerner.beadledom.correlation.CorrelationIdContext.CorrelationIdCloseable;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+class CorrelationIdContextTest {
```
venki09, line 11: Any reason to not use scala like what we have done for other modules?

correlation/src/test/java/com/cerner/beadledom/correlation/CorrelationIdContextTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+  }
+
+  @Test
+  void setCloseable() {
+  }
+
+  @Test
+  void resetCorrelationId() {
+  }
```
venki09, line 129: Do these need to filled out?

client/resteasy-client/src/main/java/com/cerner/beadledom/client/resteasy/ResteasyCorrelationIdContext.java
```diff
@@ -27,9 +27,13 @@
 
   @Override
   public String getCorrelationId() {
-    HttpHeaders headers = ResteasyProviderFactory.getContextData(HttpHeaders.class);
 
-    String correlationId;
+    String correlationId = com.cerner.beadledom.correlation.CorrelationIdContext.get();
```
venki09, line 31: Where does the setting of correlationID happen? Is it on the server-side?

correlation/pom.xml
```diff
@@ -0,0 +1,39 @@
...
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-engine</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-log4j12</artifactId>
```
johnlcox, line 35: Any reason to not use logback for the tests?

correlation/src/main/java/com/cerner/beadledom/correlation/CorrelationIdContext.java
```diff
@@ -0,0 +1,68 @@
...
+  private static final ThreadLocal<String> correlationId = new ThreadLocal<>();
+
+  private CorrelationIdContext() {
+  }
+
+  /**
+   * Returns the current Correlation ID, {@code null} if no Correlation ID is in context.
+   */
+  @Nullable
+  public static String get() {
```
johnlcox, line 21: I was surprised to see all these methods are static. In other cases where we have used `ThreadLocal` we have encapsulated it into instances. Is there a reason for not having usages of this class operate on an instance?

nathanschile: I originally did use instances, https://github.com/cerner/beadledom/blob/520830961d11c6e79c18896eee02f5f45f557dc7/correlation/src/main/java/com/cerner/beadledom/correlation/ThreadLocalCorrelationContext.java and ended up switching to static methods for simplicity for the consumer. Consumers may want to add additional functionality to the implementation, which would be afforded if it was a instance. I'll work on switching it back.

correlation/src/test/java/com/cerner/beadledom/correlation/CorrelationIdContextTest.java
nathanschile: Yeah, static methods and try-with-resources. I also wanted to give junit 5 a try.

correlation/src/main/java/com/cerner/beadledom/correlation/CorrelationIdContext.java
nathanschile: Updated CorrelationIdContext to be a interface and created a implementation. Is that what you were thinking?

https://github.com/cerner/beadledom/pull/180/commits/8a7b63eb74bda01165675f0a39c69f30d5ab6f9d

correlation/src/test/java/com/cerner/beadledom/correlation/CorrelationIdContextTest.java
nathanschile: Tested earlier in the tests, removed https://github.com/cerner/beadledom/pull/180/commits/8a7b63eb74bda01165675f0a39c69f30d5ab6f9d

correlation/pom.xml
nathanschile: Actually, I don't need slf4j, removed, d8bbd750

client/resteasy-client/src/main/java/com/cerner/beadledom/client/resteasy/ResteasyCorrelationIdContext.java
nathanschile: Yes, server-side in the CorrelationIdFilter, https://github.com/cerner/beadledom/blob/8a7b63eb74bda01165675f0a39c69f30d5ab6f9d/jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/CorrelationIdFilter.java#L82

correlation/src/test/java/com/cerner/beadledom/correlation/CorrelationIdContextTest.java
johnlcox: I like that we can keep a similar BDD-style output with junit5.

correlation/src/main/java/com/cerner/beadledom/correlation/CorrelationIdContext.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * A context that provides access to the Correlation ID.
+ */
+public interface CorrelationIdContext {
+  /**
+   * Returns the current Correlation ID, {@code null} if no Correlation ID is in context.
+   */
+  @Nullable
+  String get();
```
johnlcox, line 14: To me, this method sort of reads like I'm getting the context rather than the ID from the context.

Could there ever be other things on the context? Would it be helpful for `get`/`set` to be `getId`/`setId`?

correlation/src/main/java/com/cerner/beadledom/correlation/CorrelationIdContext.java
```diff
@@ -0,0 +1,51 @@
...
+   */
+  void set(@Nullable String correlationId);
+
+  /**
+   * Sets the Correlation ID. If {@code correlationId} is null, a Correlation ID is generated. The
+   * returned {@link Closeable} object will clear the Correlation ID when {@link Closeable#close()}
+   * is called. This method is useful to use with try-with-resources when setting a Correlation ID
+   * for the duration of the try block.
+   */
+  CorrelationIdCloseable setCloseable(@Nullable String correlationId);
```
johnlcox, line 27: I'm not following the purpose of this method. Would this be better served with a method like `withId` that took a lambda?

correlation/src/main/java/com/cerner/beadledom/correlation/CorrelationIdContext.java
nathanschile: "ever"..maybe, but for right now, I just see us having the ID. I'm good with renaming it to getId/setId.

correlation/src/main/java/com/cerner/beadledom/correlation/CorrelationIdContext.java
nathanschile: Yeah, it would work better. db80eb4b

correlation/src/main/java/com/cerner/beadledom/correlation/CorrelationContextImpl.java
```diff
@@ -0,0 +1,32 @@
+package com.cerner.beadledom.correlation;
+
+import java.util.UUID;
+import javax.annotation.Nullable;
+
+/**
+ * Default implementation that stores the correlation context in {@link ThreadLocal}.
+ */
+public class CorrelationContextImpl implements CorrelationContext {
```
johnlcox, line 9: Do you think we need a separate interface and default implementation at this time? Would it be worth deferring that decision by making `CorrelationContext` be a class marked `final` so it's the only implementation, but can't be extended leaving us room to split it later if there is demand?

johnlcox: This might also require a private constructor with a factory method for creation as well since splitting later would result in moving the constructor to the implementation.

nathanschile: I can do that. I was thinking that people may want to eventually use https://github.com/open-telemetry/opentelemetry-java/tree/master/api/src/main/java/io/opentelemetry/correlationcontext

johnlcox: Yeah, I could see that. Another option instead of what I suggested above could be to name this `ThreadLocalCorrelationContext`.

johnlcox: Either way I think we can implement this library in such a way that we are open to implementing some of this with opentelemetry in the future as a passive change.

nathanschile: Renamed to `ThreadLocalCorrelationContext` c1774fd5

correlation/src/main/java/com/cerner/beadledom/correlation/CorrelationContext.java
```diff
@@ -0,0 +1,43 @@
...
+   */
+  default <V> V withId(@Nullable String correlationId, Callable<V> task) throws Exception {
+    try {
+      setId(correlationId);
+      return task.call();
+    } finally {
+      clearId();
+    }
+  }
+
```
johnlcox, line 41: nit: Extra line breaks.

client/resteasy-client/src/main/java/com/cerner/beadledom/client/resteasy/BeadledomResteasyClientBuilder.java
```diff
@@ -300,7 +301,7 @@ public BeadledomResteasyClient build() {
     CorrelationIdContext correlationIdContext =
         new ResteasyCorrelationIdContext(
             clientConfig.correlationIdName(),
-            clientConfig.correlationIdName());
+            clientConfig.correlationIdName(), new ThreadLocalCorrelationContext());
```
johnlcox, line 304: This will work since the `ThreadLocal` in the implementation is static, but do we have any thoughts on how we can keep this working if we were to use an implementation using opentelemetry? (e.g. how do we make sure the container implementation and the client implementation within that container match?)

nathanschile: I probably should expose setting the correlationcontext on the builder object. But to your question, the consumer would need to bind to the same implementation. We could also use a CorrelationContextManager that could lookup the implementation based on SPI.

correlation/src/main/java/com/cerner/beadledom/correlation/CorrelationContextImpl.java
nathanschile: Switched to be a final class CorrelationContext, c1f9800a

correlation/src/main/java/com/cerner/beadledom/correlation/CorrelationContext.java
nathanschile: Removed, c1f9800a

docs/source/manual/correlation.rst
```diff
@@ -0,0 +1,59 @@
...
+  </dependencies>
+
+Usage
+-----
+
+Setting the Correlation ID
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: java
+  CorrelationContext correlationContext = new ThreadLocalCorrelationContext();
```
johnlcox, line 33: These need to be updated with your latest change.

correlation/src/test/java/com/cerner/beadledom/correlation/CorrelationContextTest.java
```diff
@@ -0,0 +1,125 @@
...
+import org.junit.jupiter.api.Test;
+
+class CorrelationContextTest {
+
+  CorrelationContext correlationContext = CorrelationContext.create();
+
+  @Nested
+  @DisplayName("#getId")
+  class Get {
+    @AfterEach
```
johnlcox, line 17: Would it work if the `clearId` stuff was in the root class? Would it still run after each test of the nested classes?

docs/source/manual/correlation.rst
nathanschile: Updated https://github.com/cerner/beadledom/pull/180/commits/fef50c057da3fd5605b5312ab6c1d4774e57cd09

correlation/src/test/java/com/cerner/beadledom/correlation/CorrelationContextTest.java
nathanschile: Yes, that works. I like it. 0770a372

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/CorrelationIdFilter.java
```diff
@@ -33,6 +33,7 @@
   private static final String DEFAULT_MDC_NAME = "Correlation-Id";
```
johnlcox, line 33: Should we change the default header name in 4.0 to match our internal usage?

nathanschile: We already do that in our internal module, CernerResteasyModule. I don't think the open-source beadledom would want the name to include `Cerner`

PR #182, opened 2020-04-17
Resteasy 3.11 uplift

venki09:

* Uplift to resteasy 3.11

### How was it tested?

> Describe the strategy you used to validate these changes. For example, this should mention if you added unit/integration tests and summarize any manual testing you performed by running your branch locally or in a dev environment.

### How to test

### DIFF
--- a/jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/exceptionmapping/JsonMappingExceptionMapperSpec.scala
+++ b/jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/exceptionmapping/JsonMappingExceptionMapperSpec.scala
@@ -51,7 +51,7 @@ class JsonMappingExceptionMapperSpec
         dispatcher.invoke(request, response)
 
         response.getStatus mustBe BAD_REQUEST.getStatusCode
-        response.getOutputHeaders.getFirst(CONTENT_TYPE) mustBe MediaType.APPLICATION_JSON
+        response.getOutputHeaders.getFirst(CONTENT_TYPE) mustBe MediaType.APPLICATION_JSON_TYPE
         response.getContentAsString must beBadRequestError("Unable to map request JSON")
       }
     }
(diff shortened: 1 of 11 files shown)


jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/exceptionmapping/JsonMappingExceptionMapperSpec.scala
```diff
@@ -51,7 +51,7 @@ class JsonMappingExceptionMapperSpec
         dispatcher.invoke(request, response)
 
         response.getStatus mustBe BAD_REQUEST.getStatusCode
-        response.getOutputHeaders.getFirst(CONTENT_TYPE) mustBe MediaType.APPLICATION_JSON
+        response.getOutputHeaders.getFirst(CONTENT_TYPE) mustBe MediaType.APPLICATION_JSON_TYPE
```
johnlcox, line 54: Is the change from `APPLICATION_JSON` to `APPLICATION_JSON_TYPE` another change that is going to break downstream consumers of resteasy?

venki09: Probably, I ll add that to the changelog entry as well.

johnlcox: I think technically this new version of Beadledom will still be backward compatible with resteasy 3.6.3.Final since we are only making this change in the tests and the other changes in this PR seem totally passive.

This means downstream consumers could upgrade Beadledom but still manage resteasy to a lower version than 3.11.

venki09: Yes thats true. c2e3df00 I have added a note in changelog to make it a little more clear.

PR #187, opened 2020-05-13
Adding HttpHealthDependency helper class

venki09:

* Adding HttpHealthDependency class to enable checking health of HTTP services.

### How was it tested?

* Tests
* Consumed it in a service locally.

### How to test

### DIFF
--- a/docs/source/manual/health.rst
+++ b/docs/source/manual/health.rst
@@ -70,3 +70,21 @@ Usage
       return "foobar";
     }
   }
+
+* If your health dependency is a HTTP service, you can use the HttpHealthDependency_ class to declare your dependency.
+
+.. code-block:: java
+
+  public class MyModule extends AbstractModule {
+    @Override
+    public void configure() {
+      install(new HealthModule());
+    }
+
+    @ProvidesIntoSet
+    HealthDependency provideMyHealthDependency() {
+      return new HttpHealthDependency("https://base-url/meta/availability", "dependencyName", isPrimary);
+    }
+  }
+
+.. _HttpHealthDependency: https://github.com/cerner/beadledom/blob/master/health/service/src/main/java/com/cerner/beadledom/health/HealthDependency.java
--- a/health/service/src/main/java/com/cerner/beadledom/health/HttpHealthDependency.java
+++ b/health/service/src/main/java/com/cerner/beadledom/health/HttpHealthDependency.java
@@ -1,0 +1,52 @@
+package com.cerner.beadledom.health;
+
+import java.io.IOException;
+import java.util.Objects;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+
+/**
+ * Implementation of {@link HealthDependency} to check the health of any HTTP service.
+ */
+public class HttpHealthDependency extends HealthDependency {
+  private final String availabilityUrl;
+  private final String name;
+  private final OkHttpClient client;
+
+  /**
+   * Constructs an instance of {@link HttpHealthDependency} with the given parameters.
+   *
+   * @param availabilityUrl String representing URL of the availability health check.
+   * @param name String representing name of the dependency
+   * @param isPrimary Boolean representing whether the dependency is primary or not.
+   */
+  public HttpHealthDependency(String availabilityUrl, String name, boolean isPrimary) {
+    super(isPrimary);
+    Objects.requireNonNull(availabilityUrl);
+    Objects.requireNonNull(name);
+
+    this.availabilityUrl = availabilityUrl;
+    this.name = name;
+    this.client = new OkHttpClient.Builder().build();
+  }
+
+  @Override
+  public HealthStatus checkAvailability() {
+    Request request = new Request.Builder().url(availabilityUrl).head().build();
+
+    try (Response response = client.newCall(request).execute()) {
+      if (response.isSuccessful()) {
+        return HealthStatus.create(200, getName() + " is available.");
+      }
+      return HealthStatus.create(503, getName() + " is not available.");
+    } catch (IOException e) {
+      return HealthStatus.create(503, getName() + " is not available.");
+    }
+  }
+
+  @Override
+  public String getName() {
+    return name;
+  }
+}
--- a/health/service/src/main/java/com/cerner/beadledom/health/HttpHealthDependency.java
+++ b/health/service/src/main/java/com/cerner/beadledom/health/HttpHealthDependency.java
@@ -8,6 +8,8 @@ import okhttp3.Response;
 
 /**
  * Implementation of {@link HealthDependency} to check the health of any HTTP service.
+ *
+ * @author Venkatesh Sridharan.
  */
 public class HttpHealthDependency extends HealthDependency {
   private final String availabilityUrl;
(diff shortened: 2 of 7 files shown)


docs/source/manual/health.rst
```diff
@@ -70,3 +70,21 @@ Usage
       return "foobar";
     }
   }
+
+* If your health dependency is a HTTP service, you can use the HttpHealthDependency_ class to declare your dependency.
```
charanpanchagnula, line 74: this is minor, but could you mention here that this comes from okhttp?

health/service/src/main/java/com/cerner/beadledom/health/HttpHealthDependency.java
```diff
@@ -0,0 +1,54 @@
...
+    Objects.requireNonNull(name);
+
+    this.availabilityUrl = availabilityUrl;
+    this.name = name;
+    this.client = new OkHttpClient.Builder().build();
+  }
+
+  @Override
+  public HealthStatus checkAvailability() {
+    Request request = new Request.Builder().url(availabilityUrl).head().build();
```
charanpanchagnula, line 38: Possibly stupid question..do we not configure retries for health checks?

docs/source/manual/health.rst
venki09: I am not sure I understand this. Are you wanting me to mention that HttpHealthDependency class comes from OkHttp?

health/service/src/main/java/com/cerner/beadledom/health/HttpHealthDependency.java
venki09: hmm, I have not seen it being done in the past. Do others have thoughts on this?

docs/source/manual/health.rst
charanpanchagnula: No i wanted to mention in passing in the doc that we're using okhttp to perform the health check because i dont think we're using okhttp for very much else in beadledom

venki09: Is that necessary though? Because thats an implementation detail

health/service/src/main/java/com/cerner/beadledom/health/HttpHealthDependency.java
```diff
@@ -0,0 +1,54 @@
...
+
+    try (Response response = client.newCall(request).execute()) {
+      if (response.isSuccessful()) {
+        return HealthStatus.create(200, getName() + " is available.");
+      }
+      return HealthStatus.create(503, getName() + " is not available.");
```
nathanschile, line 44: probably use the real status from the call and not 503.

docs/source/manual/health.rst
johnlcox: I agree that how we are making the HTTP calls is an implementation detail and doesn't seem necessary to explain in the docs.

health/service/src/main/java/com/cerner/beadledom/health/HttpHealthDependency.java
charanpanchagnula: Will the consuming class for this change its logic in anyway if this is a 4xx instead of a 503?

nathanschile: If they choose to. Internally, on the server side, we check if it is a 2XX status [1]. Either way, it is more appropriate to return the actual status, 

```java
  /**
   * Returns the HTTP status code returned by the dependency, as seen from the location of the
   * health check resource (eg. `200`).
   *
   * <p>For dependencies without an HTTP health check, this should simulate what such a health check
   * would return - e.g. 200 for healthy, 503 for unavailable.
   */
  public abstract int getStatus();
```

[1] https://github.com/cerner/beadledom/blob/master/health/service/src/main/java/com/cerner/beadledom/health/internal/HealthChecker.java#L155

health/service/src/main/java/com/cerner/beadledom/health/HttpHealthDependency.java
nathanschile: OkHttp already defaults to retry on connection errors. I think that is all that would be required for health checks.

johnlcox: I don't think the Ruby one does retries so I think that's the way to go here for now. Especially considering the architectures group discussions around potentially completely revamping how we do health checks.

health/service/src/main/java/com/cerner/beadledom/health/HttpHealthDependency.java
```diff
@@ -0,0 +1,54 @@
...
+   * @param isPrimary Boolean representing whether the dependency is primary or not.
+   */
+  public HttpHealthDependency(String availabilityUrl, String name, boolean isPrimary) {
+    super(isPrimary);
+    Objects.requireNonNull(availabilityUrl);
+    Objects.requireNonNull(name);
+
+    this.availabilityUrl = availabilityUrl;
+    this.name = name;
+    this.client = new OkHttpClient.Builder().build();
```
nathanschile, line 33: From the OkHttpClient docs,

```
OkHttp performs best when you create a single `OkHttpClient` instance and reuse it for all of your HTTP calls.
```

Lets update the constructor to take in a OkHttpClient. Doing this would also allow consumer to configure additional interceptors.

health/service/src/main/java/com/cerner/beadledom/health/HttpHealthDependency.java
johnlcox: The RFC for the dependencies endpoint says this too "SHOULD simply relay HTTP status of dependencies basic availability check"

health/service/src/main/java/com/cerner/beadledom/health/HttpHealthDependency.java
johnlcox: I wonder if something like this is an exclusion to that. We aren't really concerned about performance here and are just choosing OkHttp as the implementation of a simple health check call.

nathanschile: Here is some addittional information, https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/#okhttpclients-should-be-shared 

> This is because each client holds its own connection pool and thread pools. Reusing connections and threads reduces latency and saves memory.

We could have two constructors, one the has the OkHttpClient provided, and one the creates one.

venki09: But shouldn't each Health dependency have its own client instance created though? Because then we would just be initializing the class once for each dependency. I also didn't want to expose that we are using OkHttp

johnlcox: @nathanschile While I do think the okhttp docs are recommending a single client instance being shared, I think their biggest recommendation is against creating a new client for every request. 

Is reusing a client instance per dependency but not across the whole service worth the tradeoff for the value we get in not exposing the implementation (namely that we can change implementations and makes it simpler to use)?

health/service/src/main/java/com/cerner/beadledom/health/HttpHealthDependency.java
venki09: 1105d152, I am still returning 503 when an IOException is thrown. Is that okay?

health/service/src/main/java/com/cerner/beadledom/health/HttpHealthDependency.java
nathanschile: I think it is fine to have it as an implementation detail, thus not exposing it. If we determine at a later date that it is a issue having a client per dependency, then we have an option to expose it at that time.

PR #188, opened 2020-05-27
Add server relative quality factor to health resources

nathanschile:

Adds [server relative quality factor](https://download.oracle.com/otn-pub/jcp/jaxrs-2_0_rev_A-mrel-eval-spec/jsr339-jaxrs-2.0-final-spec.pdf#section.3.5) to the health resources. Having this set removes the following message from being logged when calling a health endpoint without setting the `Accept` header.

```
WARN  o.jboss.resteasy.resteasy_jaxrs.i18n - RESTEASY002142: Multiple resource methods match request "GET /meta/health/". Selecting one. Matching methods: [public abstract javax.ws.rs.core.Response com.cerner.beadledom.health.api.HealthResource.getPrimaryHealthCheckHtml(), public abstract javax.ws.rs.core.Response com.cerner.beadledom.health.api.HealthResource.getPrimaryHealthCheck()]
```

### How to test

### DIFF
--- a/health/api/src/main/java/com/cerner/beadledom/health/api/DependenciesResource.java
+++ b/health/api/src/main/java/com/cerner/beadledom/health/api/DependenciesResource.java
@@ -12,7 +12,6 @@ import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
-import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.StreamingOutput;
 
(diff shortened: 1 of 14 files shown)


health/api/src/main/java/com/cerner/beadledom/health/api/DependenciesResource.java
```diff
@@ -45,12 +44,12 @@
       @io.swagger.annotations.ApiResponse(code = 200, message = "healthy",
           response = HealthDependencyDto.class)})
   @GET
-  @Produces(MediaType.APPLICATION_JSON)
+  @Produces("application/json; qs=0.9")
```
johnlcox, line 47: Any thoughts on a good way to create a single source of truth of the 2 supported media types for health check endpoints?

The only idea occurring to me is to define them as static fields in a private class in this same package.

nathanschile: That's all I could also come up with. Created MediaType class, 98f5dc0f

PR #189, opened 2020-06-22
4.x Release Preperation

nathanschile:

* Uplifted dependencies
* Updated CHANGELOG.md

### How to test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Breaking Changes
 * Removed beadledom-swagger1 and beadledom-stagemonitor modules.
+* Changed AnnotatedObjectMapperProvider and ObjectMapperProvider to use [PropertyNamingStrategy.LOWER_CAMEL_CASE](https://javadoc.io/static/com.fasterxml.jackson.core/jackson-databind/2.10.4/com/fasterxml/jackson/databind/PropertyNamingStrategy.html#LOWER_CAMEL_CASE) from [PropertyNamingStrategy.CAMEL_CASE_TO_LOWER_CASE_WITH_UNDERSCORES](https://javadoc.io/static/com.fasterxml.jackson.core/jackson-databind/2.10.4/com/fasterxml/jackson/databind/PropertyNamingStrategy.html#CAMEL_CASE_TO_LOWER_CASE_WITH_UNDERSCORES).
 
 ### Changed
 * Upgraded RESTEasy to 4.5.5.Final.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 [Migration guide](http://engineering.cerner.com/beadledom/4.0/docs/releases/Beadledom40.html) to Beadledom 4.0 from Beadledom 3.x.
 
 ### Breaking Changes
-* Removed beadledom-swagger1 and beadledom-stagemonitor modules
+* Removed beadledom-swagger1 and beadledom-stagemonitor modules.
 
 ### Changed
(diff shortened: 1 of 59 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,17 @@
 # Beadledom Changelog
```
nathanschile, line 1: The beadledom-client projects are deprecated, should those be removed in the 4.0 release, or at a later date?

johnlcox: Probably later.

jackson/src/main/java/com/cerner/beadledom/jackson/ObjectMapperProvider.java
```diff
@@ -46,7 +46,7 @@ public ObjectMapper get() {
     // Sets default values before looking at bound features
     objectMapper.setSerializationInclusion(JsonInclude.Include.NON_ABSENT);
     objectMapper.setPropertyNamingStrategy(
-        PropertyNamingStrategy.CAMEL_CASE_TO_LOWER_CASE_WITH_UNDERSCORES);
+        PropertyNamingStrategy.SNAKE_CASE);
```
johnlcox, line 49: Should we make this `LOWER_CAMEL_CASE` for 4.0 and note it in a section about breaking changes and the migration guide? 

I'm not even sure why this is `SNAKE_CASE` ,since we generally add `@JsonProperty("camelCaseName")` with camel case names, unless it's just been this way since way before we even open sourced this and we've kept it for passivity.

CHANGELOG.md
```diff
@@ -12,12 +23,12 @@
 * The `beadledom-correlation` module.
 
 ### Changed
-* Updated resteasy to 4.x. Following changes are affected if your service is also uplifting to use 4.x version of Resteasy.
-  * Make sure to install the [Beadledom JAX-RS CorrelationIdFilter](jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/CorrelationIdFilter.java) on the server side so that correlationIds are forwarded properly when consuming a beadledom client.
+* Updated resteasy to 3.11. Following changes are affected if your service is also uplifting to use 4.x version of Resteasy.
```
venki09, line 26: This should still be resteasy 4.x correct?

jackson/src/main/java/com/cerner/beadledom/jackson/ObjectMapperProvider.java
nathanschile: There may be some old users that are depending on it to be snake case. I'm fine with making a braking change since everything new should be lower camel case.

CHANGELOG.md
nathanschile: no. In 3.6 we uplifted to 3.11.

venki09: yeah but in 4.x branch, we uplifted to resteasy4.x so the changelog for beadledom-4 should say that we uplifted to resteasy to 4.x correct?

nathanschile: Correct, Under `4.0 - In Development` I have
`Upgraded RESTEasy to 4.5.5.Final.`

Is there something else that needs to change?

venki09: I think thats all we need. Thanks!

jackson/src/main/java/com/cerner/beadledom/jackson/ObjectMapperProvider.java
nathanschile: Changed to LOWER_CAMEL_CASE, aec475c4

PR #198, opened 2021-02-25
malformed uri

kyleroush:

> Explain the changes included in this request and the reasons for them.

I have added a new jaxrs filter that will check to see if the request is a correctly formatted URI and if it is not it will short circuit the request and return a 400 status code.
I have also updated the JSON field filtering logic to only apply the filter when the request is successful. 

### How was it tested?

> I integrated these changes into a service I own and tested that when a 400 was responded instead of a 500

### How to test

### DIFF
--- a/jaxrs/src/main/java/com/cerner/beadledom/jaxrs/JaxRsModule.java
+++ b/jaxrs/src/main/java/com/cerner/beadledom/jaxrs/JaxRsModule.java
@@ -1,7 +1,9 @@
 package com.cerner.beadledom.jaxrs;
 
 import com.cerner.beadledom.jaxrs.provider.CorrelationIdFilter;
+import com.cerner.beadledom.jaxrs.provider.MalformedRequestFilter;
 import com.google.inject.AbstractModule;
+import com.google.inject.Provides;
 import javax.inject.Singleton;
 
 /**
@@ -19,5 +21,10 @@ public class JaxRsModule extends AbstractModule {
   protected void configure() {
     bind(CorrelationIdFilter.class).toProvider(CorrelationIdFilterProvider.class)
         .in(Singleton.class);
+  }
+
+  @Provides
+  MalformedRequestFilter provideMalformedRequestFilter() {
+    return new MalformedRequestFilter();
   }
 }
--- a/jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/FilteringJacksonJsonProvider.java
+++ b/jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/FilteringJacksonJsonProvider.java
@@ -11,6 +11,7 @@ import java.io.OutputStream;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
 import javax.inject.Inject;
+import javax.servlet.http.HttpServletResponse;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
@@ -32,6 +33,9 @@ public class FilteringJacksonJsonProvider extends JacksonJsonProvider {
   @Context
   UriInfo uriInfo;
 
+  @Context
+  HttpServletResponse httpServletResponse;
+
   /**
    * Creates a new instance of {@link FilteringJacksonJsonProvider}.
    */
@@ -52,6 +56,12 @@ public class FilteringJacksonJsonProvider extends JacksonJsonProvider {
   public void writeTo(
       Object o, Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType,
       MultivaluedMap<String, Object> httpHeaders, OutputStream os) throws IOException {
+        
+    if (httpServletResponse.getStatus() >= 400 ) {
+      super.writeTo(o, type, genericType, annotations, mediaType, httpHeaders, os);
+      return;
+    }
+
     String fields = uriInfo.getQueryParameters() == null ? null
         : uriInfo.getQueryParameters().getFirst("fields");
 
--- a/jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/MalformedRequestFilter.java
+++ b/jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/MalformedRequestFilter.java
@@ -12,16 +12,10 @@ import javax.ws.rs.core.Response;
 import javax.ws.rs.ext.Provider;
 
 /**
- * The ForwardedHeaderFilter reads the request context and determines if the client request was made
- * from a secure protocol (HTTPS). If the request originated from secure context, it updates the
- * request to reflect the original protocol.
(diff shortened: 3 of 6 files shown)


jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/MalformedRequestFilter.java
```diff
@@ -0,0 +1,37 @@
...
+import javax.ws.rs.ext.Provider;
+
+/**
+ * The MalformedRequestFilter reads the request context and determines if the request has a valid uri structure if it
+ * does not then it aborts the request and responds with a 400
+ *
+ * @author Kyle Roush
+ */
+@Provider
+@Priority(Priorities.AUTHENTICATION)
```
nathanschile, line 21: Could we lower this to a USER priority? Probably Not. I see that ForwardedHeaderFilter also uses UriInfo.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/MalformedRequestFilter.java
```diff
@@ -0,0 +1,37 @@
...
+@PreMatching
+public class MalformedRequestFilter implements ContainerRequestFilter {
+
+  @Override
+  public void filter(ContainerRequestContext requestContext) {
+    try {
+      URI.create(requestContext.getUriInfo().getAbsolutePath().toString());
+    } catch (IllegalArgumentException e) {
+      requestContext.abortWith(
```
nathanschile, line 30: Let's log the exception.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/FilteringJacksonJsonProvider.java
```diff
@@ -52,6 +56,12 @@ public long getSize(
   public void writeTo(
       Object o, Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType,
       MultivaluedMap<String, Object> httpHeaders, OutputStream os) throws IOException {
+        
+    if (httpServletResponse.getStatus() >= 400 ) {
```
nathanschile, line 60: Instead of pulling in the servlet-api, can we update the code to something like

```
    String fields = null;
    try {
      fields = uriInfo.getQueryParameters() == null ? null
          : uriInfo.getQueryParameters().getFirst("fields");
    } catch (Throwable e) {
      // Nothing to do. URI does not conform to grammar of RFC 2396.
    }
```

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/JaxRsModule.java
```diff
@@ -20,4 +22,9 @@ protected void configure() {
     bind(CorrelationIdFilter.class).toProvider(CorrelationIdFilterProvider.class)
         .in(Singleton.class);
   }
+
+  @Provides
```
nathanschile, line 26: Add `@Singleton`

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/JaxRsModule.java
```diff
@@ -20,4 +22,9 @@ protected void configure() {
     bind(CorrelationIdFilter.class).toProvider(CorrelationIdFilterProvider.class)
```
nathanschile, line 22: Update the class Javadoc to indicate that the module now provides the MalformedRequestFilter

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/MalformedRequestFilter.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+@Provider
+@Priority(Priorities.AUTHENTICATION)
+@PreMatching
+public class MalformedRequestFilter implements ContainerRequestFilter {
+
+  @Override
+  public void filter(ContainerRequestContext requestContext) {
+    try {
+      URI.create(requestContext.getUriInfo().getAbsolutePath().toString());
```
nathanschile, line 28: You can do `requestContext.getUriInfo().getRequestUri();` so it doesn't create a new URI.

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/provider/MalformedRequestFilterSpec.scala
```diff
@@ -0,0 +1,54 @@
+package com.cerner.beadledom.jaxrs.provider
+
+import java.net.URI
+
+import com.cerner.beadledom.json.common.model.JsonError
+import javax.ws.rs.container.ContainerRequestContext
+import javax.ws.rs.core.{MediaType, Response, UriInfo}
+import org.jboss.resteasy.specimpl.ResteasyUriInfo
+import org.mockito
+import org.mockito.{ArgumentCaptor, Mockito}
```
nathanschile, line 10: Cleanup imports

jaxrs/src/test/scala/com/cerner/beadledom/jaxrs/provider/MalformedRequestFilterSpec.scala
```diff
@@ -0,0 +1,54 @@
...
+    val containerRequestContext = Mockito.mock(classOf[ContainerRequestContext])
+    val uriInfo = new ResteasyUriInfo("http://localhost:8080", "")
+    Mockito.when(containerRequestContext.getUriInfo).thenReturn(uriInfo)
+
+    val malformedRequestFilter = new MalformedRequestFilter()
+
+    it("does not call abortWith") {
+      malformedRequestFilter.filter(containerRequestContext)
+
+      Mockito.verify(containerRequestContext).getUriInfo();
```
nathanschile, line 33: nit. This can be `verify(containerRequestContext).getUriInfo`, since Scala doesn't require the parens and semicolon. You can also `import org.mockito.Mockito.{verify, when}` so you don't need to have Mockito.

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/JaxRsModule.java
```diff
@@ -1,7 +1,9 @@
 package com.cerner.beadledom.jaxrs;
 
 import com.cerner.beadledom.jaxrs.provider.CorrelationIdFilter;
+import com.cerner.beadledom.jaxrs.provider.MalformedRequestFilter;
```
nathanschile, line 4: Update CHANGELOG.md with the added class.

nathanschile (changes requested):

jaxrs/src/main/java/com/cerner/beadledom/jaxrs/provider/MalformedRequestFilter.java
nab0310: Does this error occur at the first call of UriInfo in the stack?

